PR #74, opened 2023-04-14
add info to ports

joamatab:
add info to ports for keeping gdsfactory compatibility

@sebastian-goeldi


sebastian-goeldi:
What kind of info do you need to dump into ports? this will break when we get this https://github.com/KLayout/klayout/issues/1341

sebastian-goeldi:
If you need gdsfactory compatibility you should be able to just inherit from this class and add it, can't you?

sebastian-goeldi:
Also I am currently working on making the CplxKCell obsolete

joamatab:
we usually use the `info` dict for simulation metadata

PR #75, opened 2023-04-14
add named_instances

joamatab:
add named_instances

useful for naming circuit elements

`coupler_input`


sebastian-goeldi:
If you want to add this, please make it so that it directly reflects in the gds :)

```inst.set_property(0, name)```

and then make it as a property on the instance class

sebastian-goeldi:
Fixed with https://github.com/gdsfactory/kfactory/pull/78

PR #78, opened 2023-04-18
Add info to ports and allow named Instances

sebastian-goeldi:
Fixes #74  #75 

@joamatab

### DIFF
--- a/docs/notebooks/01_references.py
+++ b/docs/notebooks/01_references.py
@@ -133,7 +133,7 @@ c2
 
 c = kf.KCell(name="reference_sample")
 w = kf.pcells.waveguide.waveguide(length=10, width=0.6, layer=c.klib.layer(1, 0))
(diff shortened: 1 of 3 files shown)


sebastian-goeldi:
Example:

```
import kfactory as kf


c = kf.KCell()


wg = kf.pcells.waveguide.waveguide(length=10, width=1, layer=c.klib.layer(1, 0))

wg1 = c << wg
wg1.name = "wg1"


wg2 = c << wg

wg2.align("o1", wg1, "o2")

c.show()
```
image

joamatab:
This is great Sebastian,

would this work with OASIS?

sebastian-goeldi:
OASIS takes strings or int keys, so it should work with OASIS yes. In OASIS it would even be possible to do `name: "wg1`

joamatab:
amazing! :)

PR #82, opened 2023-04-28
Gdsfactory cells

joamatab:
- rename autocell to pcell (makes more sense)
- add compass pcell


sgoeldi-psiq:
any reason for not using a `klayout.db.Box` instead of a compass? In KLayout there is really no reason to use a compass, imo.

sebastian-goeldi:
Why not use `kdb.Box` or `kdb.DBox` instead of a compass? It is much leaner and gives you everything you want from a compass.

sebastian-goeldi:
w.r.t. to the renaming to pcell. I think this would be super confusing with this [guy](https://www.klayout.de/doc-qt5/code/class_PCellDeclarationHelper.html). It is in fact not one of those, nor should it ever be.

sebastian-goeldi:
And why add `if __name__ ...` blocks to cells? if you want something like that, write a test ;)

joamatab:
`cell` sounds also good to me,

PR #87, opened 2023-05-04
Add cells diff tests

SkandanC:
Add geometry diff tests for the cells. As of now, the scripts work, but we need a nice  way to store the reference gds files for the difftests. gitignore ignores all gds files.

Any thoughts on storing gds files? Maybe in gdslib as a separate branch?

@joamatab 
@sebastian-goeldi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.tar
 *.zip
 .idea
+gds/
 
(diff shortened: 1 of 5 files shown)


joamatab:
How about uploading the GDS as artifacts?

i pushed a demo

PR #105, opened 2023-05-12
rename align -> transport

sebastian-goeldi:


sebastian-goeldi:
@joamatab I think we can't do the `inst.ports.o1` in the way you want. We for sure can make it work, but autocomplete won't work afaik. We'd overwrite the `__getattr__` of `Ports` and `InstancePorts`, but nevertheless it won't autocomplete as autocompletes can't do this dynamic lookup, because they would need to run the code to know what items are available.

joamatab (approved):

gdspaul:
Request to slow this one down!  transport() was suggested 13 hours ago.  I think it deserves the same scrutiny given to all the other carefully-considered options.

PR #107, opened 2023-05-15
PDKEnclosures

sebastian-goeldi:
Allows to use a colllection of Enclosures to be applied to a cell

### DIFF
--- a/docs/layers.py
+++ b/docs/layers.py
@@ -6,4 +6,4 @@ class LAYER(kf.LayerEnum):
     SIEXCLUDE = (1, 1)
 
 
-si_enc = kf.utils.Enclosure([(LAYER.SIEXCLUDE, 2000)])
+si_enc = kf.utils.LayerEnclosure([(LAYER.SIEXCLUDE, 2000)])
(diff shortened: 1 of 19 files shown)


sebastian-goeldi:
@joamatab any ideas for names on the enclosure set naming? I am not sure what to name it. It allows you to define a set of enclosures and they can be applied to a cell and will automatically create all the enclosures defined in it.

joamatab:
i like the name  `Pdk.enclosures` 

`special_enclosures` is a bit wordy :)

sebastian-goeldi:
I think I didn't explain properly:

So, with the new one, you have one default one that depending on what "main_layers" of each enclosure are present, will apply any enclosure in it. Special enclosures would be if you need a non-standard one, so I think it would be fine to have a longer name. I mainly meant for the new class (currently named `FullEnclosure`).

The standard Enclosure collection would be in `Pdk.enclosure`

sebastian-goeldi:
* renamed `Enclosure` -> `LayerEnclosure`
* renamed `PdkEnclosure` -> `CellEnclousre`
* renamed `Pdk.enclosures` -> `Pdk.layer_enclosures`
* added `Pdk.enclosure` which is the CellEnclosure to use for the Pdk

TODO:

`PortHoles` (carved out in enclosures), are not selective for enclosures from their ports, yet.

PR #111, opened 2023-05-19
rename align

joamatab:
renames align to chain

the main issue with [align](https://gdsfactory.github.io/gdsfactory/notebooks/04_components_pack.html#align) is that it means something else in gdsfactory and is going to make users quite confused


sebastian-goeldi:
I don´t  fully agree, since one is on the equivalent of saying that `KCell.transform` is doing something, therefore `Instance.transform` cannot exist.

We could enhance align to not only allow ports to be linked but also `inst.align(kdb.Point(x1,y1), inst2, kdb.Point(x2,y2)` and also allow to do it with edges and full transformations (which would be almost ports, but not fully).

gdspaul:
> I don´t fully agree, since one is on the equivalent of saying that `KCell.transform` is doing something, therefore `Instance.transform` cannot exist. The current align is on Instance, whereas gdsfactory align is on a KCell/Component

Request for clarification—can you elaborate, what is equivalent to "saying that `KCell.transform` is doing something" and what is, uh, the other one?

joamatab:
How about the following names?

- connect (with big disclaimer that this is not merge)
- chain
- abut
- tranport
- transform_port

sebastian-goeldi:
```
Request for clarification—can you elaborate, what is equivalent to "saying that KCell.transform is doing something" and what is, uh, the other one?
```

`KCell.transform` moves all the shapes & instances in itself (but not itself, because that is not possible in GDS), and the instance moves itself. I would argue for align it is the same, one moves elements of itself and the other moves itself.

But, alas, I am not super attached to the naming. My personal list in descending liking would be:

1. transport / teleport
2. align
3. connect with the heavy disclaimer & clear distinction to netlist behavior
4. abut

joamatab:
i see, i renamed to `connect` 

let me know if the description is clear now

PR #112, opened 2023-05-21
use jupytext to pair notebooks

joamatab:


docs/notebooks/01_references.md
```diff
@@ -35,6 +35,8 @@ You can start by making a blank `KFactory` and add a single polygon to it.
 ```python
 import kfactory as kf
 
+kf.config.display_type = "image"
+
```
sebastian-goeldi, line 39: Why change to image here? The docs will already run with this behavior. If you want to reproduces locally use `export KFACTORY_DISPLAY_TYPE="image"`

.pre-commit-config.yaml
```diff
@@ -146,6 +146,11 @@ repos:
   #   rev: 2b2f0c74acdb3de316e23ceb7dd0d7945c354050
   #   hooks:
   #     - id: reorder-python-imports
+  - repo: https://github.com/mwouts/jupytext
+    rev: 79351dfac9fde7c72cc73d553f662b4887bc63f5  # CURRENT_TAG/COMMIT_HASH
+    hooks:
+    - id: jupytext
+      args: [--from, py, --to, md]
```
sebastian-goeldi, line 153: What's the purpose of this?  I don't want duplicates of py -> md in the repo. If it is simply about checking whether it is possible, the runner can take care of this already, no?

docs/star.md
sebastian-goeldi, line 1: Why add the `.md` files? They were python files intentionally.

sebastian-goeldi (changes requested):

joamatab:
I see, the mkdocs was not working for me before adding the jupytext,

you are right, having `py` and `md` files does not make too much sense and is confusing

PR #148, opened 2023-06-05
rename theta to angle

joamatab:
- renames theta to angle

### DIFF
--- a/docs/source/notebooks/01_references.py
+++ b/docs/source/notebooks/01_references.py
@@ -272,8 +272,8 @@ c
 
 # %%
 c = kf.KCell()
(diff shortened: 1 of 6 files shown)


joamatab:
two advantages:

- angle is more explicit than theta
- in gdsfactory we use angle, so it's easier not to add new concepts

what do you think?

sebastian-goeldi:
I don't want to constantly break the API and rewrite it ;). I am fine with doing it once more, but then we need to do other stuff as well. E.g. waveguide is called straight in gdsfactory, which I guess makes sense.

If we constantly break the API, I always have to halt our internal projects and backpatch them and then tell everybody to update kfactory.

PR #177, opened 2023-09-05
Layerstack

drinwater:
Add layerstack for kplugins simulations and 3d rendering 

@sebastian-goeldi 
@joamatab


tests/conftest.py
```diff
@@ -127,7 +129,19 @@ def cells(bend90, bend180, bend90_euler, taper, straight) -> list[kf.KCell]:
...
+            thickness=220,
+            zmin=0,
+            material="si",
+            info=kf.kcell.Info(mesh_order=1),
+        ),
+        clad=kf.kcell.LayerLevel(
+            layer=LAYER_CLASS.WGCLAD, thickness=3000, zmin=220, material="sio2"
+        ),
+    )
+    kcl = kf.KCLayout("Test_PDK", layer_stack=layerstack)
```
sebastian-goeldi, line 144: This won't work, at least not the way you think it would work. You can't reference layers while the KCLayout isn't constructed yet, so anything using layers must be added post-creation.

src/kfactory/kcell.py
```diff
@@ -1484,6 +1484,317 @@ class Constants(BaseSettings):
...
+zmin_heater = 1.1 / 0.001
+zmin_metal1 = 1.1 / 0.001
+thickness_metal1 = 700
+zmin_metal2 = 2.3 / 0.001
+thickness_metal2 = 700
+zmin_metal3 = 3.2 / 0.001
+thickness_metal3 = 2000
+substrate_thickness = 10.0 / 0.001
+box_thickness = 3.0 / 0.001
+undercut_thickness = 5.0 / 0.001
```
sebastian-goeldi, line 1609: Why is there stuff defined outside the class? And can we remove all the commented stuff below?

src/kfactory/kcell.py
```diff
@@ -1484,6 +1484,317 @@ class Constants(BaseSettings):
...
+            if isinstance(val.layer, LayerEnum):
+                self.layers[field].layer = (val.layer[0], val.layer[1])
+
+    def get_layer_to_thickness(self) -> dict[tuple[int, int] | LayerEnum, int]:
+        """Returns layer tuple to thickness (um)."""
+        return {
+            level.layer: level.thickness
+            for level in self.layers.values()
+            if level.thickness
+        }
```
sebastian-goeldi, line 1553: Wouldn't it make more sense to have a layer thickness in µm instead of dbu? The concept of dbu doesn´t  really apply to the z-axis. If it's in µm, it definitely should be `float` not `int`

sebastian-goeldi (changes requested):

tests/conftest.py
sebastian-goeldi: You can do this, but then it needs the logic to carry over layers from base_kcl to kcl and then apply the changes to the layerstack as well

src/kfactory/kcell.py
sebastian-goeldi: Makes sense. But I would lean towards, don't do that here. If the dbu is bigger, it could still mean you have someayers that are very thin and can't be expressed in dbu

drinwater: ah ok

tests/conftest.py
drinwater: @sebastian-goeldi could you take a loo at this? Thanks

sebastian-goeldi (approved):

PR #179, opened 2023-09-18
Add Port Connectivity checks

sebastian-goeldi:


src/kfactory/kcell.py
```diff
@@ -1464,6 +1464,7 @@ def connectivity_check(
         db: rdb.ReportDatabase | None = None,
         recursive: bool = True,
         add_cell_ports: bool = False,
+        check_layerconnectivity: bool = True,
```
joamatab, line 1467: shouldn't be `check_layer_connectivity`?

sebastian-goeldi: should be outdated already, but yes

PR #192, opened 2023-10-06
Add port center setter getter

joamatab:


src/kfactory/kcell.py
```diff
@@ -428,10 +428,7 @@ def __init__(
         self._settings: KCellSettings = KCellSettings()
         self.info: Info = Info()
         self._locked = False
-        if name is None:
-            _name = "Unnamed_!"
-        else:
-            _name = name
+        _name = "Unnamed_!" if name is None else name
```
sebastian-goeldi, line 431: Please don't touch. This is a very sensitive piece of code that I debugged more times than I want to admit

src/kfactory/kcell.py
```diff
@@ -1315,6 +1312,16 @@ def ymax(self) -> int:
...
+    @property
+    def ysize(self) -> int:
+        """Returns the x-coordinate of the left edge of the bounding box."""
+        return self._kdb_cell.bbox().top - self._kdb_cell.bbox().bottom
+
+    @property
+    def xsize(self) -> int:
+        """Returns the x-coordinate of the left edge of the bounding box."""
+        return self._kdb_cell.bbox().left - self._kdb_cell.bbox().right
+
```
sebastian-goeldi, line 1324: bbox has width() and height(), no need to do the math yourself. Also, is it really necessary to alias all of these? Because bbox has the nicety you can do `c.bbox(LAYER.WG).width()` to get the width on WG only
or we build it in directly?

```python
    @property
    def xsize(self, layer: LayerEnum|int|None = None) -> int:
        """Returns the x-coordinate of the left edge of the bounding box."""
        if layer:
            return self._kdb_cell.bbox(layer).width()
        else:
            return self._kdb_cell.bbox().width()
```

src/kfactory/kcell.py
```diff
@@ -3347,6 +3364,16 @@ def ymax(self, __val: int) -> None:
...
+    @property
+    def ysize(self) -> int:
+        """Returns the x-coordinate of the left edge of the bounding box."""
+        return self._instance.bbox().top - self._instance.bbox().bottom
+
+    @property
+    def xsize(self) -> int:
+        """Returns the x-coordinate of the left edge of the bounding box."""
+        return self._instance.bbox().left - self._instance.bbox().right
+
```
sebastian-goeldi, line 3376: see above wrt width/height

src/kfactory/kcell.py
```diff
@@ -3373,9 +3400,12 @@ def center(self) -> kdb.Point:
...
 
     @center.setter
-    def center(self, __val: kdb.Point) -> None:
+    def center(self, val: tuple[int, int] | kdb.Point) -> None:
         """Moves the instance so that the bbox's center coordinate."""
-        self.transform(kdb.Trans(__val - self.bbox().center()))
+        if isinstance(val, tuple):
+            self.transform(kdb.Trans(kdb.Point(val[0], val[1]) - self.bbox().center()))
+        else:
+            self.transform(kdb.Trans(val - self.bbox().center()))
```
sebastian-goeldi, line 3408: I would swap this:

```python
        if isinstance(val, (kdb.Point, kdb.Vector)):
            self.transform(kdb.Trans(val - self.bbox().center()))
        else:
            center = self.bbox().center()
            self.transform(kdb.Trans(val[0] - center.x, val[1] - center.y ))
```

sebastian-goeldi (changes requested):

src/kfactory/kcell.py
joamatab: sure :)

src/kfactory/kcell.py
joamatab: great!

src/kfactory/kcell.py
joamatab: thank you

src/kfactory/kcell.py
joamatab: done

sebastian-goeldi (approved):

PR #195, opened 2023-10-07
rename autorename_ports to auto_rename_ports

joamatab:
- rename autorename_ports to auto_rename_ports 
- add UMInstance center, x, y getter and setter
- add default name to Instance by using parent_cell_name_x_y
 fixes https://github.com/gdsfactory/kfactory/issues/197

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -962,7 +962,7 @@ class KCell:
             h.update(_hash)
         return h.digest()
 
-    def autorename_ports(self, rename_func: Callable[..., None] | None = None) -> None:
+    def auto_rename_ports(self, rename_func: Callable[..., None] | None = None) -> None:
         """Rename the ports with the schema angle -> "NSWE" and sort by x and y.
 
         Args:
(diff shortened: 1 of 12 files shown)


src/kfactory/kcell.py
```diff
@@ -3000,7 +3000,11 @@ def __getattr__(self, name):  # type: ignore[no-untyped-def]
     def name(self) -> str | None:
         """Name of instance in GDS."""
         prop = self.property(PROPID.NAME)
-        return str(prop) if prop is not None else None
+        return (
+            str(prop)
+            if prop is not None
+            else f"{self.kcl[self.cell_index].name}_{self.x}_{self.y}"
+        )
```
sebastian-goeldi, line 3007: Why return a name if it's not a named instance? This is a bit weird, because it doesn't necessarily reflect what there is and also it's not necessarily unique. I can create you two connected waveguides that will return the same instance name, but are properly connected waveguides.

sebastian-goeldi (changes requested):

src/kfactory/kcell.py
joamatab: otherwise if a component accepts an instance, then `clean_value(instance)` will return None and it will fail when trying to replace the invalid characters `,|.|?`
I think giving them default names is also what we did in gdsfactory, it makes it nicer for circuit simulations

sebastian-goeldi (approved):

PR #220, opened 2023-11-01
better_odd_port_width_message

### LINKED ISSUE #219: better error for port non divideable by 2
```python

import kfactory as kf 

c = kf.KCell()
# this should fail
c.create_port(name='A', width=475, center=(0, 0), layer=0, angle=0)

# this should give you a better error message
orientation = 0
center = (0, 0)
trans = kf.kdb.DCplxTrans(1, orientation, False, center[0], center[1])
c.create_port(name='B', dwidth=0.475, center=(0, 0), layer=0, dcplx_trans=trans)

```

joamatab:
closes #219

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -23,7 +23,7 @@ from hashlib import sha3_512
 from pathlib import Path
 from tempfile import gettempdir
 from types import ModuleType
-from typing import Any, Literal, TypeAlias, TypeVar, overload
+from typing import Any, Literal, TypeAlias, TypeVar, cast, overload
 
 import cachetools.func
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/kfactory/kcell.py
```diff
@@ -3916,7 +3919,15 @@ def create_port(
                 kcl=self.kcl,
             )
         elif dcplx_trans is not None:
-            assert dwidth is not None
+            if dwidth is None or dwidth <= 0:
+                raise ValueError("dwidth needs to be set")
+            elif dwidth is not None and dwidth > 0:
+                width = round(dwidth / self.kcl.dbu)
+                width = cast(int, width)
```
sebastian-goeldi, line 3926: I believe this is not necessary. This only has an impact on the type checking. width should already be an integer due to the `round`. Does mypy fail without this? Also afaik this might wipe width because cast does a nop in no-type-checking context

sebastian-goeldi (changes requested):

sebastian-goeldi (approved):

PR #228, opened 2023-11-14
Virtual Cells

sebastian-goeldi:


joamatab:
awesome!

some docs on how to use would be great :)

sebastian-goeldi:
It's still a draft. I am currently writing the tests.

sebastian-goeldi:
@joamatab @tvt173 please take a look whether this is something we could work with for all-angle routing. an example is in tests/test_virtual.py as the last test.

src/kfactory/kcell.py
```diff
@@ -4581,12 +4631,576 @@ def _check_cell_ports(p1: Port, p2: Port) -> int:
...
+    _settings: KCellSettings
+    info: Info
+    kcl: KCLayout
+    insts: list[VInstance]
+    _name: str | None
+
+    def __init__(
+        self,
+        name: str | None = None,
+        kcl: KCLayout = kcl,
```
joamatab, line 4649: Why not do the same trick that we do in KCell?

src/kfactory/kcell.py
```diff
@@ -4581,12 +4631,576 @@ def _check_cell_ports(p1: Port, p2: Port) -> int:
     return check_int
 
 
+class VKCell(BaseModel, arbitrary_types_allowed=True):
```
joamatab, line 4634: Why not inherit from KCell?

src/kfactory/routing/aa/optical.py
joamatab, line 1: maybe add docs and test for this

joamatab (commented):
Looks good, 

is the idea to use virtual cells for non-manhattan components?

can we mix and match both?

sebastian-goeldi:
Yes, you can (kinda). You can use KCells in VInstances. But to get VInstance into a KCell, you need to do `vinst.insert_into(kcell,...)`. But as long as it's a virtual instance, it can use either of them. But a KCell can never contain any virtual component, they'll become real on insert

src/kfactory/kcell.py
sebastian-goeldi: You mean about the name? Because we don't have an index here ;) Nor does it really serve a lot of purpose here. We can have VKCells without names, it will only cause problems if we want to insert it unnamed as a KCell. That case is caught further down.

src/kfactory/kcell.py
sebastian-goeldi: Because it is not attached to a KCell, nor do we really want everything from a KCell. We only want a slim subset and more importantly, it needs to not be attached to a layout object

src/kfactory/routing/aa/optical.py
sebastian-goeldi: There are already tests in `test_virtual.py`.

Docs, yeah could, not super happy with it yet.

tests/test_data/ref/BendCircular_W1_R10_LWG_EWGSTD_A180_AS1.gds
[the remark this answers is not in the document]
sebastian-goeldi: yep

demo.gds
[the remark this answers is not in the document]
sebastian-goeldi: nope. artifact from docs. added to gitignore

Issue #53, opened 2023-02-24
allow usage in klayout embedded python


proppy:
Would re-using `kfactory` component model when implementing klayout pcells for interactive usage (inside klayout gui) be within the scope of the project?

Context is the discussion about a recent refactoring of open source PDKs pcells in https://github.com/efabless/globalfoundries-pdk-libs-gf180mcu_fd_pr/issues/28

A few question toward this:
- does kfactory aim to feature parity with gdsfactory drawing primitive (`rectangle`, `boolean` ops, etc)
- would `KCell` just work on top of `pya` instead of `klayout.db/lay` https://github.com/gdsfactory/kfactory/blob/1bcdc7912e8c00613e8e2bcdc74cd24deef6bd11/src/kfactory/__init__.py#L6-L7
- would it makes sense to package an lite version of kfactory with less dependencies (no yaml, no scipy)

Let me know what you think!

sebastian-goeldi:
I think it is within the scope that we can make this happen. I think the main issue being the imports. But I can see that with the following solution:

```python
try:
    import klayout.dbcore as kdb
    import klayout.lay as lay
except ImportError:
    logger.warning("Cannot find klayout modules dbcore and lay, trying to import klayout packages")
else:
    import pya as kdb
    import pya as lay
```
> does kfactory aim to feature parity with gdsfactory drawing primitive (rectangle, boolean ops, etc)
Kind of yes. But they are a bit more "advanced" as in they are built-in to klayout directly. (In the following assume `c` (and `c1, c2,...`) is an instance of `kf.KCell` and `layer_index` is either a `kf.LayerEnum` or `int`)
* rectangle:
   ```python
   c.shapes(layer_index).insert(kf.kdb.Box(left, bottom, right, top)```
* booleans:
   I would argue klayout booleans are just straight up better than gdsfactory booleans
   ```python
   region = kf.kdb.Region(c1.begin_shapes_rec(layer_index)) # get all shapes (recursively from cell c1)
   region2 = kf.kdb.Region()
   region2.insert(kf.kdb.Box(0, 500, 1000, -500))
   # can do this an arbitrary amount of times

   # boolean operations
   sub_region = region - region2 # r NOT r2
   add_region = region + region2 # r OR r2
   and_region = region & region2 # r AND r2
   xor_region = region ^ region2 # r XOR r2```
   
> would `KCell` just work on top of `pya` instead of `klayout.db/lay`
Yes, pya contains both and more (such as Qt objects)

[...]

PR #230, opened 2023-12-14
allow gui mode in kfactory

joamatab:
allow using GUI mode in klayout, you just have to make sure you don't install klayout

it would be great to support PCells this way

similar to the functionality we had with [flayout](https://github.com/flaport/flayout)

fixes https://github.com/gdsfactory/kfactory/issues/53

@proppy
@flaport
@sebastian-goeldi 
@SkandanC

### DIFF
--- a/src/kfactory/__init__.py
+++ b/src/kfactory/__init__.py
@@ -9,6 +9,7 @@ Uses the klayout package as a backend.
 
 try:
     import pya
+
     kdb = pya
     lay = pya
     rdb = pya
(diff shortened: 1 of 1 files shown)


sebastian-goeldi:
It's not that simple unfortunately. Just because we are using "native" klayout, it's not just done with enabling loading if pya is available.  You'd have to make KCell compatible with [`PCellDeclarationHelper`](https://klayout.de/doc/code/class_PCellDeclarationHelper.html) . So, unless we do that, there is very little use in just enabling loading kfactory into KLayout.

PR #231, opened 2023-12-29
allow tuples in info and settings

joamatab:
- allow tuples for info and settings
- add get method so we can use a default value


src/kfactory/kcell.py
```diff
@@ -163,26 +163,30 @@ class KCellSettings(BaseModel, extra="allow", validate_assignment=True, frozen=T
...
-            if not isinstance(value, str | int | float | SerializableShape):
+            if not isinstance(value, str | int | float | SerializableShape | tuple):
                 data[name] = str(value)
         return data
 
     def __getitem__(self, key: str) -> Any:
         return getattr(self, key)
 
+    def get(self, __key: str, default: Any | None = None) -> Any:
+        return getattr(self, __key) if hasattr(self, __key) else default
```
sebastian-goeldi, line 176: Fine by me, but `getattr` supports a default value by design already. Also, `Any | None`is redundant as `Any` includes `None` afaik

sebastian-goeldi (changes requested):

sebastian-goeldi:
Can you also change tuple to `Sequence`, as far as I understand, klayout supports lists.

Also, can you add to the read/write test some cases for this?

src/kfactory/kcell.py
```diff
@@ -163,26 +163,30 @@ class KCellSettings(BaseModel, extra="allow", validate_assignment=True, frozen=T
     @model_validator(mode="before")
     def restrict_types(
         cls, data: dict[str, Any]
-    ) -> dict[str, int | float | SerializableShape | str]:
+    ) -> dict[str, int | float | SerializableShape | str | tuple[float | int, ...]]:
```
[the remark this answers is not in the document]
joamatab: I fixed it :) thank you

src/kfactory/kcell.py
joamatab: I fixed it :) thank you

joamatab:
i added a test for it as well

thank you Sebastian

sebastian-goeldi (approved):

sebastian-goeldi:
@joamatab any idea why the pre-commit fails? I am only on my phone, so I was not able to find the problem from the action.

Also, can you confirm `isinstance(Sequence)` works for list and tuples? I wasn't aware that works.


Otherwise looks good to me. Feel free to merge if you get the pre-commit to work.

src/kfactory/kcell.py
```diff
@@ -163,35 +164,42 @@ class KCellSettings(BaseModel, extra="allow", validate_assignment=True, frozen=T
     @model_validator(mode="before")
     def restrict_types(
         cls, data: dict[str, Any]
-    ) -> dict[str, int | float | SerializableShape | str]:
+    ) -> dict[str, int | float | SerializableShape | Sequence | str]:
```
sebastian-goeldi, line 167: Ah, now I know what the problem is, the type should be:

`Sequence[int|float|SerializableShape|str]`

We could also leave the SerializableShape away, as I don't see much use case for that.

sebastian-goeldi (changes requested):

src/kfactory/kcell.py
joamatab: do we want to allow bool as well?

sebastian-goeldi: Yes, it's supported by klayout, so we can add it.

joamatab: When we do this 

`Sequence[int|float|SerializableShape|str]` i get pytest errors

and if we don't do it we get mypy errors :)

sebastian-goeldi: works locally on 3.12 for me and in cicd. Probably time for updating pytest and mypy with `python -m pip install -e .[dev] -U --upgrade-strategy eager` ;)

sebastian-goeldi (approved):

sebastian-goeldi:
@joamatab you can merge now if you are ok with the changes I made

joamatab:
Awesome, thank you Sebastian

Issue #236, opened 2024-01-07
'UMKCell' object has no attribute 'x'


joamatab:
![image](https://github.com/gdsfactory/kfactory/assets/4514346/e73e7f53-ccd4-4e73-89dd-3c5310e22aea)

PR #237, opened 2024-01-07
add x,y center for Cell and UMKCell

joamatab:
fixes https://github.com/gdsfactory/kfactory/issues/236


src/kfactory/kcell.py
```diff
@@ -2967,6 +2977,22 @@ def ysize(self) -> float:
...
+    @property
+    def y(self) -> float:
+        """Y coordinate of the port in um."""
+        return self.parent._kdb_cell.dbbox().center().y
+
+    @property
+    def center(self) -> tuple[float, float]:
+        """Coordinate of the port in um."""
+        vec = self.parent._kdb_cell.dbbox().center()
+        return (vec.x, vec.y)
```
sebastian-goeldi, line 2994: to keep this consistent with other classes, just return the DPoint instead of the tuple, this is what we did in other cases. We could theoretically rethink this wrt the new Polygon handling, but I think return a point is safer, as it indicates the relation to dbu as well

src/kfactory/kcell.py
```diff
@@ -1363,6 +1363,16 @@ def get_meta_data(self) -> None:
...
+    @property
+    def x(self) -> int:
+        """Returns the x-coordinate of the center of the bounding box."""
+        return self._kdb_cell.bbox().center().x
+
+    @property
+    def y(self) -> int:
+        """Returns the y-coordinate of the center of the bounding box."""
+        return self._kdb_cell.bbox().center().y
+
```
sebastian-goeldi, line 1375: do we have center here?

sebastian-goeldi (changes requested):

src/kfactory/kcell.py
sebastian-goeldi, line 2994:
```suggestion
    def center(self) -> tuple[float, float]:
        """Coordinate of the port in um."""
        return self.parent._kdb_cell.dbbox().center()
```

Other center functions are returning the DPoint, only the port isn't doing that, maybe we should change it there as well

src/kfactory/kcell.py
joamatab: i think so, otherwise the test wouldn't pass

sebastian-goeldi (approved):

sebastian-goeldi:
@joamatab can you make a newsfragment? Or I can also add it

src/kfactory/kcell.py
joamatab: center was already in KCell.center if that's what you meant

sebastian-goeldi: Sounds good

PR #238, opened 2024-01-10
some adjustments for layerstack

sebastian-goeldi:
Some Suggestions for layerstack adjustments.


sebastian-goeldi:
@drinwater @SkandanC  I made some changes which will play nicer with type checkers and the like and also to simplify the return types. In general I think we should avoid having `tuple[int,int] | LayerEnum`. It should just return the common type, so I rewrote the `LayerLevel` to convert `LayerEnum` in the init to `(layerenum.layer, layerenum.datatype)`

PR #249, opened 2024-02-12
Allow inheritance from standard cells

sebastian-goeldi:
Move the code creating the `KCell` from `__call__` for `_kcell`.

This will allow to make inheritance.

E.g.

```python
```Create a post-processed waveguide```
import kfactory as kf

class TestBendEuler(kf.cells.euler.BendEuler):  # type: ignore[unused-ignore, misc]
    def __init__(self) -> None:
        super().__init__(kf.kcl)

    @kf.cell  # type: ignore[misc, unused-ignore]
    def __call__(self, width: float) -> kf.KCell:  # type: ignore[override, unused-ignore]
        c = self._kcell(
            width=width,
            radius=30,
            layer=LAYER.WG,
            enclosure=wg_enc,
            angle=90,
            resolution=150,
        )
        c.info.creation_time = "2023-02-12Z23:00:00"

        return c

# could also use bend_euler = TestBendEuler()
bend = TestBendEuler()(width=1)

```


sebastian-goeldi:
@joamatab

joamatab:
not really, this seems overly complicated

sebastian-goeldi:
You have to choose, either make it inheritable as a class or use no decorator.

With this you now have three choices:

- The above
- `my_kcell.dup()` (needs name change to avoid collision, i.e. doesn't change the cached KCell)
- use e.g. `kf.cells.straight.straight._kcell(...)` to get the uncached version directly and wrap that in a `@cell` decorator to cache (this will not set all settings though, only the ones passed to the new function with `@cell`)

I can't give you something to inject a post_process into an existing function. At that point, just unlock the cell and modify it, it's the same as if you were grabbing kwargs from the footprint of the function.

joamatab:
then, l prefer keeping the cell decorator as it is and then passing an explicit `post_process` function


https://github.com/gdsfactory/gdsfactory/pull/2524

PR #250, opened 2024-02-12
GitHub actions fix for release-drafter

sebastian-goeldi:


sebastian-goeldi:
@joamatab this change might be good for gdsfactory too. this will ensure the require-label only runs after the autolabeler

PR #252, opened 2024-02-18
Add port access for instance arrays and pretty print for ports

### LINKED ISSUE #203: support ports in a instance array
How can we support ports in an instance array?

sebastian-goeldi:
Fixes #203 

Usage:

```python
inst = c.creat_inst(..., a=kf.kdb.Vector(x1, y1), b=kf.kdb.Vector(x2, y2), na=4, nb=5)
inst[port_name, 3, 5]
```


sebastian-goeldi:
@joamatab I added `print` to ports and port, what do you think? This would replace pprint_ports of gdsfactory

PR #271, opened 2024-03-06
add update to info

joamatab:


sebastian-goeldi:
What's the reason for needing this?

joamatab:
we do a lot of this in gdsfactory 

```python
c.info.update(**info_settings)
```

sebastian-goeldi:
Then you only need the dict and not the info object one, right?

That makes this way easier. If you don't need iterable and Info object, I don't see why we should add it here. It just makes the signature unreadable.

joamatab:
Yes, good point

src/kfactory/kcell.py
```diff
@@ -220,6 +220,19 @@ def __setitem__(
...
+                self.__setitem__(key, value)
+        elif isinstance(data, Info):
+            for key, value in data.model_dump().items():
+                self.__setitem__(key, value)
+        elif isinstance(data, Iterable):
+            for key, value in data:
+                self.__setitem__(key, value)
+        else:
+            raise TypeError("Unsupported data type for update")
+
```
sebastian-goeldi, line 235:
```suggestion
    def update(self, data: dict[str, bool | int | float | str]) -> None:
        for key, value in data.items():
            settatr(self, key, value)
        
    def __iadd__(self, other: Info):
        for key, value in other.items():
            setattr(self, key, value)
        for key, value in other.model_dump().items():
            setattr(self, key, value)
```

joamatab:
Looks good to me,

feel free to merge

joamatab:
included in this one

https://github.com/gdsfactory/kfactory/pull/273

PR #287, opened 2024-04-05
return uminstance with moving references

joamatab:
allows to do 

```
c = gf.Component()
ref = c << gf.c.straight()
ref.d.movex(10).d.rotate(90)
```


sebastian-goeldi:
What's the reason not to return the UMInstance?

Wouldn't that make more sense because you can do:

```python
inst.d.rotate(90).movex(15.5)
```
And so on

sebastian-goeldi (approved):

PR #293, opened 2024-04-10
Allow dicts in info

joamatab:
related to
https://github.com/gdsfactory/gdsfactory/pull/2626

it even works serializing dicts to klayout

image

@thomasdorch


src/kfactory/kcell.py
```diff
@@ -202,7 +210,7 @@ def restrict_types(
         data: dict[str, MetaData],
     ) -> dict[str, MetaData]:
         for name, value in data.items():
-            if not isinstance(value, str | int | float | Sequence):
+            if not isinstance(value, str | int | float | Sequence | dict | None):
```
sebastian-goeldi, line 213: If you only change it here, it will absolutely break mypy in any code that actually uses this. And a general dict is definitely not ok. If you e.g. feed a function here it won't be deserialized correctly

sebastian-goeldi (changes requested):

PR #295, opened 2024-04-14
Add smart routing

sebastian-goeldi:
Allow general bundle routing

image

@joamatab 

TODO:
- add optical router parts
- test and aggregate collisions

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: "v4.5.0"
+    rev: "v4.6.0"
     hooks:
       - id: check-added-large-files
       - id: check-case-conflict
@@ -32,7 +32,7 @@ repos:
         args: ['--keep']
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.3.2
+    rev: v0.4.1
     hooks:
       # Run the linter.
       - id: ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ authors = [
     {name = "gdsfactory community", email = "contact@gdsfactory.com"},
 ]
 dependencies = [
-  "klayout >= 0.28.17",
+  "klayout >= 0.29.0",
   "scipy",
   "ruamel.yaml",
   "cachetools >= 5.2.0",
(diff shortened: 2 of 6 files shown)


sebastian-goeldi:
(Current) usage:
```python
from functools import partial

from random import randint

import kfactory as kf

ps: list[kf.Port] = []
pe: list[kf.Port] = []

c = kf.KCell()

for angle in [-1, 0, 1, 2]:
    for xi in range(5):
        ts = (
            kf.kdb.Trans(angle, False, 0, 0)
            * kf.kdb.Trans(100_000, -50_000)
            * kf.kdb.Trans(randint(0, 100), randint(0, 100) + xi * 10_000)
        )
        ps.append(
            c.create_port(
                name="i" + str((angle + 1) * 5 + xi),
                trans=ts,
                width=1000,
                layer=kf.kcl.layout.layer(1, 0),
            )
        )
        te = kf.kdb.Trans(2, False, 250_000, 50_000 * angle) * kf.kdb.Trans(
            randint(0, 100), randint(0, 100) - xi * 10_000
        )
        pe.append(
            c.create_port(
                name="o" + str((angle + 1) * 5 + xi),
                trans=te,
                width=1000,
                layer=kf.kcl.layout.layer(1, 0),
            )
        )


b = kf.cells.circular.bend_circular(width=1, radius=10, layer=kf.kcl.layer(1, 0))
s = partial(kf.cells.straight.straight_dbu, width=1000, layer=kf.kcl.layer(1, 0))

routers = kf.routing.manhattan.route_smart(
    start_ports=ps, end_ports=pe, bend90_radius=10_000, separation=2000
)

for p1, p2, router in zip(ps, pe, routers):
    kf.routing.optical.place90(
        c, p1=p1, p2=p2, pts=router.start.pts, straight_factory=s, bend90_cell=b
    )

c.show()
```

joamatab:
i tested it in a few cases and found some issues

image

sebastian-goeldi:
Can you show the whole thing here and what you were trying to connect?

Depending on that, You might need to change the call a bit. This is not necessarily intended to just allow you to connect all components (because it makes strong assumptions about the order of ports) and the target needs to be a bank per bundle. If start and endports per bundle overlap, it will likely cause issues.

sebastian-goeldi:
Also collision detection isn't built-in yet. That's the next step.

sebastian-goeldi:
Collision detection:

image


By default kfactory will raise a RuntimeError and send the cell plus lyrdb with errors to klive. Detection is on all port layers or they can be defined in `route_bundle`

PR #301, opened 2024-04-22
Add indirect routing of endpoints

joamatab:


joamatab:
works great, can we merge it?

sebastian-goeldi:
image

Not for all cases yes ;)

sebastian-goeldi:
Nvm, just fixed it

image


Give me 10 min, then we can merge it

sebastian-goeldi:
It works under  the assumption that all the ports that will meet are one cell/bounding box. The router routes to the bounding box of all starting ports.

E.g. in the following it will create the bounding box of all the ports and then route to the top most.

image


This might not be desirable in all cases but is a fair assumption I think. Otherwise the router will not know where it is allowed to route. The current assumption is:

- in front of the port it's free until the router meets the bounding box (which can grow when other routers are doing anything)

PR #306, opened 2024-05-02
`pprint_ports` add missing type switch

joamatab:


sebastian-goeldi:
why though? just pass `type="um"` to make it use that type by default?

PR #312, opened 2024-05-11
better serializer

joamatab:


src/kfactory/kcell.py
```diff
@@ -2803,6 +2803,7 @@ def wrapped_cell(
                     for key, value in params.items():
                         if isinstance(value, frozenset):
                             params[key] = fs2d(value)
+                    print(f"Creating cell {f.__name__!r} with {key=} {params}\n")
```
sebastian-goeldi, line 2806: I suppose this is not really necessary in the production code?

src/kfactory/kcell.py
```diff
@@ -6616,6 +6618,8 @@ def clean_value(
         return "_".join(clean_value(v) for v in value)
     elif isinstance(value, dict):
         return dict2name(**value)
+    elif isinstance(value, BaseModel):
+        return clean_value(value.model_dump())
     elif hasattr(value, "name"):
         return clean_name(value.name)
     elif (
```
sebastian-goeldi, line 6625: Why not prefer just using me over serializing the whole basemodel?

src/kfactory/kcell.py
```diff
@@ -2711,7 +2711,7 @@ def cell(
         check_instances: bool = True,
         snap_ports: bool = True,
         add_port_layers: bool = True,
-        cache: Cache[int, Any] | dict[int, Any] | None = None,
+        cache: Cache[int, Any] | dict[int, Any] = {},
```
sebastian-goeldi, line 2714: Why change the cache? Also this is very dangerous, my guess is after this there is exactly one global cache. I don't think that's advisable.

src/kfactory/enclosure.py
```diff
@@ -580,9 +581,15 @@ def __init__(
...
-            (str(self), self.main_layer, tuple(list(self.layer_sections.items())))
+        hasher = hashlib.sha256()
+        # Create a consistent string representation of the object
+        state_str = (
+            f"{str(self)} {self.main_layer} {sorted(self.layer_sections.items())}"
         )
+        # Update the hasher with the string encoded to bytes
+        hasher.update(state_str.encode("utf-8"))
+        # Convert the hash to an integer
+        return int.from_bytes(hasher.digest(), byteorder="big")
```
sebastian-goeldi, line 592: What's the advantage of using this over the `hash` ?

src/kfactory/kcell.py
```diff
@@ -6576,6 +6582,14 @@ def dict2name(prefix: str | None = None, **kwargs: dict[str, Any]) -> str:
     return clean_name(_label)
 
 
+def dict2hash(prefix: str | None = None, **kwargs: dict[str, Any]) -> int:
+    """Returns hash from a dict."""
+    hasher = hashlib.sha256()
+    state_str = dict2name(prefix, **kwargs)
+    hasher.update(state_str.encode("utf-8"))
+    return int.from_bytes(hasher.digest(), byteorder="big")
```
sebastian-goeldi, line 6590: why use this  slow way instead of properly hashing structs?

src/kfactory/kcell.py
```diff
@@ -6616,6 +6630,8 @@ def clean_value(
         return "_".join(clean_value(v) for v in value)
     elif isinstance(value, dict):
         return dict2name(**value)
+    elif isinstance(value, BaseModel):
+        return str(hash(value))
```
sebastian-goeldi, line 6634: BaseModel has no proper implementation of `__hash__` afaik. So if you were to do:

```python
@kf.cell
def test_cell(model: MyModel) -> kf.KCell:
    c = kf.KCell
    return c

c1 = test_cell(my_model)
c2 = test_cell(my_model.model_copy())

kf.kcl.write(...)
```
will error out because hash(my_model) != hash(my_model.model_copy())

sebastian-goeldi (changes requested):

PR #328, opened 2024-05-23
Keep mirror flag

joamatab:
keep mirror flag in instances if they have been mirrored before


sebastian-goeldi:
This doesn't do anything?

PR #339, opened 2024-05-26
Change to  `d{key}` instead of `d.{key}`

sebastian-goeldi:
Drop UMKCell, UMInstance and UMPort and add properties/functions based in um to KCell/Instance/Port


sebastian-goeldi:
If we merge this, we will need to bump to at least `0.15`

PR #345, opened 2024-05-29
fix global config settings only being set on first import

sebastian-goeldi:


src/kfactory/kcell.py
```diff
@@ -5664,6 +5681,16 @@ def connect(
             use_mirror: If False mirror flag does not get applied from the connection.
             use_angle: If False the angle does not get applied from the connection.
         """
+        if allow_width_mismatch is None:
```
fp-matt, line 5684: Should be `allow_layer_mismatch`?

src/kfactory/kcell.py
```diff
@@ -4498,6 +4505,16 @@ def connect(
...
```
fp-matt, line 4508: should be `allow_layer_mismatch`?

PR #346, opened 2024-05-29
Add hash to long automatic cell name

joamatab:
Add a 8 char hash at the end of the cell name if it's too long


src/kfactory/kcell.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import annotations
 
 import functools
+import hashlib
```
sebastian-goeldi, line 14: We are already importing hashlib further down once. Albeit differently, but I also don't know what we would use hashlib otherwise for except shas.

sebastian-goeldi (changes requested):

sebastian-goeldi (approved):

PR #362, opened 2024-05-31
Add `overwrite_existing` and `layout_cache` options to the cell decorator

sebastian-goeldi:
- `KCLayout.cell(overwrite_existing)` will make sure that when we are creating a new cell any pre-existing cell with the same name will be deleted.
  WARNING: This might cause unwanted sideeffects like suddenly deleted references etc if an overwrite was performed, as all those cells will be deleted and therefore any instance of them as well.
- `KCLayout.cell(layout_cache)`  will use the current KCLayout as its cache. This means if a cell is already existant in the layout, this will use that cell instead of creating one.


joamatab (approved):
looks great!

PR #375, opened 2024-06-04
Drop rec_dicts flag and make it default to use rec dict inspection

joamatab:
@sebastian-goeldi


sebastian-goeldi:
No. This will slow it down considerably.

Best I can give you is a config flag.

tvt173:
> No. This will slow it down considerably.

Have you tried profiling or is this an assumption? I would challenge that assumption. Certainly it will slow it down, but I doubt it will be that significant.

> And what sense does it make to enable it by default?
> 
> Just enable it in cells where it makes sense?

It makes it much easier to work with if you can just always do it, rather than the user needing to know they need to set that flag in the decorator of cells which take dictionary arguments... and being confused why it works sometimes but not all the time

joamatab:
sounds good,

can we at least change the name?

i like
`allow_dicts`

PR #400, opened 2024-06-15
fix basename usage in cell decorator

flaport:


sebastian-goeldi:
Changing to a `factory_name` property after discussion with @flaport

flaport:
Looks good to me 👍

PR #405, opened 2024-06-20
Use tempfile if called from IPython

doddgray:
This PR is meant to address https://github.com/gdsfactory/gdsfactory/issues/2782. 

It modifies `kcell` to handle the case when it is called from an IPython kernel. Currently an IPython call causes a bug when `inspect` is used to identify the calling python file. This just adds a check to see if the "filename" looks like an IPython call and if so streams to KLayout via a temporary OAS file (in `Path(tempfile.gettempdir()) / "gdsfactory"`) rather than the `build` directory in the calling file's repository.

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -7597,7 +7597,6 @@ def show(
     _kcl_paths: list[dict[str, str]] = []
 
     if isinstance(layout, KCLayout):
-        # print("\tCase 1: type(layout)==KCLayout")
         file: Path | None = None
         spec = importlib.util.find_spec("git")
         if spec is not None:
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -7580,8 +7580,7 @@ def show(
         frame = stk[2]
         frame_filename_stem = Path(frame.filename).stem
         if frame_filename_stem.startswith("<ipython-input"):    # IPython Case
-            ipython_shell_id = frame_filename_stem.rstrip(">").split("-")[-1]
-            name = f"ipython-{ipython_shell_id}"
+            name = "ipython"
         else:                                                   # Normal Python kernel case
             if frame.function != "<module>":
                 name = frame_filename_stem + "_" + frame.function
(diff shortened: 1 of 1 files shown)


src/kfactory/kcell.py
```diff
@@ -7656,7 +7662,10 @@ def show(
             else:
                 wtd = repo.working_tree_dir
                 if wtd is not None:
-                    root = Path(wtd) / "build/gds"
+                    if name.startswith("ipython-"):         # IPython Case: use tempfile
+                        root = Path(tempfile.gettempdir()) / "gdsfactory"
```
sebastian-goeldi, line 7666: We aren't gdsfactory here ;)

Jokes aside, what do you think of instead using the build-dir (if available) and naming it `ipython.oas` (no lernelid etc)?

doddgray: I don't have a very strong feeling about using the local repo's 'build' directory (as opposed to a temp directory) for saving OAS files generated from IPython calls. My thinking was that 
 (1) oasis files streamed to KLayout from interactive python sessions are likely not meant to be saved, so use a temp directory and let the OS clean them up
 (2) I often have more than one ipython kernel running concurrently (VSCode interactive sessions for different test files) and I wanted to avoid them overwriting each other's Oasis files. Hence tagging the file with some IPython related ID. 

That said I don't think the directory matters much and I don't know if it matters if multiple kernels are overwriting each others' OASIS files once they are already open in KLayout and thus loaded in memory. I would rather conform to whatever conventions work with the rest of the package code. I'll merge in recent changes if necessary and make these changes to the PR the later this morning.

doddgray:
@sebastian-goeldi  how does this look? I simplified the changes to reflect your suggestion. LMK if this needs anything else to merge.

sebastian-goeldi:
LMK

Let me remove that one commented line then I'll merge. I'll try to release soon. There are some annoying manhattan routing bugs that currently are blocking for gdsfactory to upgrade. Once I figure those out we should be good for a new release

src/kfactory/kcell.py
sebastian-goeldi: Sounds good to me. I also don't have strong opinions on which name to use. The name only has implications on whether klive will open a new layout view or not. I would stick to one  name (as we have now). This ensures that even though the files are not really ephemeral they are at least somewhat limited (i.e. shell gets (ideally) one file, ipython gets one and otherwise one per python file with `__main__`)

doddgray:
I will make that last edit in ~3 hours.

doddgray:
I removed the commented print statement (line ~7600) that I think you were referencing. LMK if anything else is needed to merge.

PR #407, opened 2024-06-23
control_cellname_length

joamatab:
fixes https://github.com/gdsfactory/gdsfactory/issues/2907

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -793,6 +793,8 @@ class KCell:
             _name = "Unnamed_!"
         else:
             _name = name
+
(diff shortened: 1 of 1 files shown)


sebastian-goeldi:
I think this is not really advisable. This could create duplicate names, and it would be very much not obvious why or where that happens. What's the motivation for this?

It will still be shortened on writing.

As well as this doesn't catch

```python
c.name = ...
```

tvt173:
i agree with @sebastian-goeldi ... let the `@cell` decorator and/or `save_layout_options` take care of it...

PR #409, opened 2024-06-25
Routing indirect fixes

sebastian-goeldi:


sebastian-goeldi:
@joamatab can you try this in gdsfactory? I think it passes all tests now unless I made a mistake

PR #417, opened 2024-07-08
fix route_length

### LINKED ISSUE #416: wrong route.length
https://github.com/gdsfactory/gdsfactory/issues/2950

joamatab:
fixes #416


sebastian-goeldi:
By this argument you would need to do the same for straight and tapers. Because this would assume that your device might override what is physical length vs path lenght and therefore you wouldn't be able to use the standard anymore. Also if you want to integrate it, at least don't just make all the tests fail. As, now, all the factories would need to incorporate this. But I heavily disagree that all bends should have this. As this is rather an aggregated data than directly. And the class is called `OpticalManhattanRoute` after all.

sebastian-goeldi:
@joamatab is the current update through #418 satisfactory?

PR #431, opened 2024-07-18
add packing functions for Instances and KCells

sebastian-goeldi:
<details>

<summary>example.py</summary>

```python
import kfactory as kf
from kfactory.packing import pack_instances, pack_kcells


@kf.cell
def box(width: int, height: int) -> kf.KCell:
    c = kf.KCell()
    c.shapes(c.kcl.layer(1, 0)).insert(kf.kdb.Box(width, height))
    return c


c = kf.KCell("Box Packing")

ig1 = pack_kcells(
    c,
    [box(w, h) for w, h in [(500, 500) for _ in range(15)] + [(15_000, 1000)]],
    max_width=16000,
    spacing=1000,
)

small_boxes = [c << box(w, h) for w, h in ((500, 500) for _ in range(15))]
bigbox = c << box(15_000, 1000)
bigbox.transform(kf.kdb.Trans.R90)
ig2 = pack_instances(
    c,
    small_boxes + [bigbox],
    max_width=16000,
    spacing=1000,
)

ig2.movex(ig1.xmax + 5000 - ig2.xmin)
c.show()
```

</details>

gives:

image

Also clean up some of the instance group code.

@joamatab @akshat-ja @gdspaul

### DIFF
--- a/src/kfactory/packing.py
+++ b/src/kfactory/packing.py
@@ -1,0 +1,79 @@
+"""Utilities for packing KCells and Instances."""
+
+from collections.abc import Sequence
+
+import rpack  # type: ignore[import-untyped,unused-ignore]
+
+from . import kdb
+from .kcell import Instance, InstanceGroup, KCell
+
+
+def pack_kcells(
+    target: KCell,
+    kcells: Sequence[KCell],
+    max_width: int | None = None,
+    max_height: int | None = None,
+    spacing: int = 0,
+) -> InstanceGroup:
+    """Pack KCells.
+
+    Args:
+        target: KCell to place the packed instances in.
+        kcells: Sequence of KCells from which to create instances.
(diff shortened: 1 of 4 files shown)


akshat-ja:
Let me test a bit more on my side @sebastian-goeldi then I'll green light.

src/kfactory/packing.py
```diff
@@ -0,0 +1,79 @@
+"""Utilities for packing KCells and Instances."""
+
+from collections.abc import Sequence
+
+import rpack  # type: ignore[import-untyped,unused-ignore]
```
joamatab, line 5: how about moving this import inside the function, so that rpack is an optional dependency?

sebastian-goeldi: And what would we gain of it not being a dependency?

PR #439, opened 2024-07-25
Improve layers so that they can be passed easily by LayerInfo and not only by index

sebastian-goeldi:
TODO:
- [x] make `Layers` better
- [x] allow dynamic creation and registering of `Layers`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ docs/modules
 # ipython
 .ipynb_checkpoints
 *.ipynb
+.virtual_documents
 
 
 demo.gds
--- a/docs/source/complex_cell.py
+++ b/docs/source/complex_cell.py
@@ -6,11 +6,11 @@
 #       extension: .py
 #       format_name: light
 #       format_version: '1.5'
-#       jupytext_version: 1.14.5
+#       jupytext_version: 1.16.2
 #   kernelspec:
-#     display_name: kernel_name
+#     display_name: Python 3 (ipykernel)
 #     language: python
-#     name: kernel_name
+#     name: python3
 # ---
 
 from layers import LAYER, si_enc
@@ -30,7 +30,7 @@ def composite_cell() -> kf.KCell:
     )
     wg = c << straight(1000, 5000, 5000)
 
-    wg.align("o1", bend, "o2")
+    wg.connect("o1", bend, "o2")
 
     c.add_port(name="1", port=bend.ports["o1"])
     c.add_port(name="2", port=wg.ports["o2"])
--- a/docs/source/layers.py
+++ b/docs/source/layers.py
@@ -16,9 +16,11 @@
 import kfactory as kf
 
 
-class LAYER(kf.LayerEnum):
-    SI = (1, 0)
-    SIEXCLUDE = (1, 1)
+class LayerInfos(kf.LayerInfos):
+    SI: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 0)
+    SIEXCLUDE: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 1)
 
 
-si_enc = kf.enclosure.LayerEnclosure([(LAYER.SIEXCLUDE, 2000)])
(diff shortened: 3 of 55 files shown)


sebastian-goeldi:
The new structure will allow this:

```python
import kfactory as kf


class Constants(kf.kcell.Constants):
    WG_WIDTH: int = 2000


class Layers(kf.kcell.LayerInfos):
    WG: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 0, "WG")
    WGEX: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 1, "WG")


kcl = kf.KCLayout[Layers, Constants]("TEST", layer_infos=Layers, constants=Constants)
c = kcl.kcell()

kcl.constants.WG_WIDTH

c.kcl.layer_infos.WG
c.kcl.layer_infos.CLAD

c.kcl.constants.WG
c.kcl.constatns.WGEX
```

Autocompletion currently only works for `kcl.constants` and `kcl.layer_infos`.
image

Mypy is able to detect missing/wrong fields though
image


Is the increased complexity worth it though? These changes will necessitate that we do

`kcl.kcell()` in the future instead of `kf.KCell(...)` (because otherwise we would need to do  `kf.KCell[MyLayerInfosClass, MyConstantsClass](kcl=kcl)`

@gdspaul @joamatab @tvt173

gdspaul:
do you mean that users who are happy to use the default `kcl` are still going to have to type `kf.kcl.kcell` to make a cell?

sebastian-goeldi:
not really, this works perfectly fine (just mypy complaining in my case):

```python
import kfactory as kf

c = kf.KCell()

c.shapes(c.kcl.layer(1, 0)).insert(kf.kdb.Box(500))
c.show()
```

gdspaul:
does the layerinfo init really need a layer name put in? it's been hard to sneak nice things past mypy here, but maybe there's a way to internally set layerinfo.name to match the layer attribute names without annoying mypy. pseudopython—
```
class Layers(baseclass):
 WG = layerinfo(1,0) # wire it up internally to become layerinfo(1,0,'WG')
```

sebastian-goeldi:
> does the layerinfo init really need a layer name put in? it's been hard to sneak nice things past mypy here, but maybe there's a way to internally set layerinfo.name to match the layer attribute names without annoying mypy. pseudopython—
> 
> ```
> class Layers(baseclass):
>  WG = layerinfo(1,0) # wire it up internally to become layerinfo(1,0,'WG')
> ```

It does not in fact, but what it needs is the type annotation due to it being a pydantic model, unfortunately.

so it's happy with

```python
class Layers(kf.kcell.LayerInfos):
    WG: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 0)


and it will make it to `WG: kf.kdb.LayerInfo = kf.kdb.LayerInfo(1, 0, "WG")` but `: kf.kdb.LayerInfo` is necessary

**Comment by sebastian-goeldi** [comment]
dataclasses are also not great if you want stuff like `asdict` to work properly, so I don't have a great solution for that.

**Comment by gdspaul** [comment]
my comment was confusing, my bad. the layers concept isn't a good fit for dataclasses either. but maybe there's *something* that could be implemented—let's call it `UserFacingLayersBaseClass`, so user code could read
```
class Layers(UserFacingLayersBaseClass):
 WG = kf.kdb.LayerInfo(1,0)
```
which does not need to be annotated, and then there's some function in the KCLayout infrastructure that does this job:
```
# user_defined_layer_class = Layers
pdk_layers: ActualLayersPydanticModel = make_pydantic_thingy_from_user_class(user_defined_layer_class)
# validation is done
pdk.layers = pdk_layers
pdk.layers.model_dump() # but Layers.model_dump() is not defined
```

**Comment by sebastian-goeldi** [comment]
I don't think that would work. I mean you are free to do any of this. You just can't pass it to the KCLayout creation.

But you can always do this if you do not want autocomplete

```python
kcl.layer_infos = kf.kcell.LayerInfos(WG=kf.kdb.LayerInfo(1,0))
```

PR #454, opened 2024-08-19
fix #452

### LINKED ISSUE #452: post_process is applied multiple times for cached cells?
**Describe the bug**
post_process for the @cell decorator appears behave differently than in gdsfactory 7. It seems to be applied multiple times for cached cells.

**To Reproduce**
The MWE here is in gdsfactory but the problem stems from kfactory:
```python
from functools import partial

import gdsfactory as gf

SOME_LAYER = (1, 0)


def add_component_name_label_on_layer(component, layer) -> None:
    component.add_label(
        component.name,
        position=component.ports[0].dcplx_trans * gf.kdb.DPoint(-0.1, 0),
        layer=layer,
    )


def test_no_duplicate_labels():

    @gf.cell(
        post_process=[partial(add_component_name_label_on_layer, layer=SOME_LAYER)]
    )
    def component_with_label(**kwargs):
        c = gf.components.straight(**kwargs)
        return c

    @gf.cell
    def container():
        c = gf.Component()
        c1 = c << component_with_label()
        c2 = c << component_with_label()
        c2.move((0, 5))
        c3 = c << component_with_label()
        c3.move((0, 10))
        return c

    c = container()
    c.show()

    assert len(c.get_labels(layer=SOME_LAYER)) == 3  # We get 9 == 3
```

**Expected behavior**
We would have 3 straights with one label each. Instead, all of the straights have 3 duplicated labels.

**Screenshots**
#### Normal view
image
#### After moving text labels that are duplicated
image

sebastian-goeldi:


nikosavola:
Does the cache work correctly when you call the cell again but without post processing or with different post processing?

sebastian-goeldi:
Since it's part of the decorator it's not really relevant for the cache as the cache also gets defined in the same go. KFactory does not have something like a global cache

PR #491, opened 2024-10-11
add instance iterator which returns coord plus port

sebastian-goeldi:
@gdspaul 

access is through:

```python
for i_a, i_b, p in inst.ports.each_by_array_coord():
     ...
```


plisdku:
beautiful. thanks @sebastian-goeldi

joamatab (approved):

PR #516, opened 2024-11-16
fix: MetaData type alias and __all__ exports

### LINKED ISSUE #515: MetaDate type error
**Describe the bug**
The `MetaData` TypeAlias in the kcell.py file fails type checking

**To Reproduce**
1. Go to 'src/kfactory/kcell.py'
2. Scroll down to 'MetaData variable initialization on line 155'
3. See type error

**Expected behavior**
There should be no type error

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Ubuntu 
 - Version 24

**Additional context**
This variable is also not exported in the __all__ variable, but its used in the [gdsfactory repo](https://github.com/gdsfactory/gdsfactory). I am working on improving type hints there, so it should be exported there too for the type checker

MatthewMckee4:
Fixes #515

## Summary by Sourcery

Fix the MetaData type alias to support recursive types and include MetaData in the module's __all__ exports.

Bug Fixes:
- Correct the MetaData type alias to properly handle recursive types.

Enhancements:
- Add MetaData to the __all__ exports in kcell.py.

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -92,6 +92,7 @@ __all__ = [
     "save_layout_options",
     "LayerEnum",
     "KCellParams",
+    "MetaData",
 ]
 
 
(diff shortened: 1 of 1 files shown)


sebastian-goeldi:
I don't think it's strictly necessary to export MetaData this way, as long as gdsfactory imports it as `from kfactory.kcell import MetaData` this should be fine and passes mypy afaik.

MatthewMckee4:
image
This is from gdsfactory.cell.

My apologies, you are right, mypy doesnt pick it up (unsure why not). Regardless of this, for a better user experience for those of use who have strict type hinting enabled in vscode for pylance, Do you think we could keep this change?

MatthewMckee4:
Thank you

sebastian-goeldi:
Yeah, as long as it's only in kcell, we can keep it

PR #528, opened 2024-12-16
Add BaseKCell class and update VKCell and KCell to inherit from it

MatthewMckee4:
## Summary by Sourcery

Introduce a new BaseKCell class to consolidate shared functionality between KCell and VKCell, and refactor these classes to inherit from BaseKCell, enhancing code organization and maintainability.

Enhancements:
- Introduce BaseKCell class to serve as a base class for KCell and VKCell, consolidating shared attributes and methods.
- Refactor KCell and VKCell to inherit from the new BaseKCell class, reducing code duplication and improving maintainability.

### DIFF
--- a/src/kfactory/kcell.py
+++ b/src/kfactory/kcell.py
@@ -29,7 +29,6 @@ from typing import (
     TYPE_CHECKING,
     Annotated,
     Any,
-    Generic,
     Literal,
     Protocol,
     TypeAlias,
@@ -597,7 +596,7 @@ class PROPID(IntEnum):
 class LockedError(AttributeError):
     """Raised when a locked cell is being modified."""
 
-    def __init__(self, kcell: BaseKCell[Any]):
+    def __init__(self, kcell: BaseKCell):
         """Throw _locked error."""
         super().__init__(
             f"KCell {kcell.name} has been locked already."
@@ -2464,7 +2463,7 @@ class VInstance(BaseModel, arbitrary_types_allowed=True):  # noqa: E999,D101
         self._ports = VInstancePorts(self)
 
     def bbox(self, layer: int | LayerEnum | None = None) -> kdb.DBox:
-        return self.cell.bbox().transformed(self.trans)  # type: ignore
+        return self.cell.bbox().transformed(self.trans)
 
     def __getitem__(self, key: int | str | None) -> Port:
         """Returns port from instance.
@@ -3127,10 +3126,7 @@ class VInstance(BaseModel, arbitrary_types_allowed=True):  # noqa: E999,D101
         self.center = val  # type: ignore[assignment]
 
 
-_TBox = TypeVar("_TBox", kdb.Box, kdb.DBox)
-
-
-class BaseKCell(BaseModel, ABC, Generic[_TBox], arbitrary_types_allowed=True):
+class BaseKCell(BaseModel, ABC, arbitrary_types_allowed=True):
     """Base class for shared attributes between VKCell and KCell."""
 
     _ports: Ports = PrivateAttr()
@@ -3161,9 +3157,6 @@ class BaseKCell(BaseModel, ABC, Generic[_TBox], arbitrary_types_allowed=True):
     @name.setter
     @abstractmethod
     def name(self, value: str) -> None: ...
-
-    @abstractmethod
-    def bbox(self) -> _TBox: ...
 
     @abstractmethod
     def dup(self) -> Self: ...
@@ -3221,7 +3214,7 @@ class BaseKCell(BaseModel, ABC, Generic[_TBox], arbitrary_types_allowed=True):
         Args:
             ports: list/tuple (anything iterable) of ports.
             prefix: string to add in front of all the port names
-            suffix: string to add the end of all the port names.
+            suffix: string to add at the end of all the port names
             keep_mirror: Keep the mirror part of the transformation of a port if
                 `True`, else set the mirror flag to `False`.
         """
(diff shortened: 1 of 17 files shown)


MatthewMckee4:
Most of the additions/deletions are moving code around (moving Instance and related classes to above class BaseKCell

sebastian-goeldi:
This PR still makes me a bit uncomfortable. I am not so sure it is a good idea to change the classes where we just get a small class out of it but it changes 12000 lines

MatthewMckee4:
That's understandable. Ive moved code back around since i didn't really need to move it in the first place.

This is my idea for the inheritance based on this PR:
```mermaid
classDiagram
    class BaseKCell {
    }

    class ComponentBase {
    }

    class KCell {
    }

    class VKCell {
    }

    class Component {
    }

    class ComponentAllAngle {
    }

    
    BaseKCell <|-- KCell
    BaseKCell <|-- ComponentBase
    BaseKCell <|-- VKCell

    ComponentBase <|-- Component
    KCell <|-- Component

    ComponentBase <|-- ComponentAllAngle
    VKCell <|-- ComponentAllAngle
```
I think will help a lot with ComponentBase class, mainly syntax highlighting, but will also allow us to remove a good few "type: ignore" statements, but i understand its a weird change so i understand if we close this.

Is there anything that can be done to make you more comfortable with this PR?

src/kfactory/kcell.py
```diff
@@ -1586,21 +1688,9 @@ def __getitem__(self, key: int | str | None) -> Port:
...
-    @property
-    def settings_units(self) -> KCellSettingsUnits:
-        """Dictionary containing the units of the settings.
-
-        Set by the [@cell][kfactory.kcell.KCLayout.cell] decorator.
-        """
-        return self._settings_units
+    def __hash__(self) -> int:
+        """Hash the KCell."""
+        return hash(self._kdb_cell)
```
sebastian-goeldi, line 1693: Here I would just return `hash((self.kcl.library.name(), self._kdb_cell.cell_index())`

I believe if you just hash the cell, you are taking the literal address of it. Which afaik is not unique, you can have multiple klayout cells pointing to the same cell, but have a different python wrapper.

src/kfactory/kcell.py
```diff
@@ -1635,9 +1725,12 @@ def ghost_cell(self, value: bool) -> None:
...
         self._kdb_cell.ghost_cell = value
 
-    def __getattr__(self, name):  # type: ignore[no-untyped-def]
+    def __getattr__(self, name: str) -> Any:
         """If KCell doesn't have an attribute, look in the KLayout Cell."""
-        return getattr(self._kdb_cell, name)
+        try:
+            return super().__getattr__(name)  # type: ignore
+        except Exception:
+            return getattr(self._kdb_cell, name)
```
sebastian-goeldi, line 1733: Are you sure this is necessary? If not, I would like to keep it as the original way. Exception handling is not super cheap (also not that expensive in the python context, but still saves 4-5 sec on the tests).

src/kfactory/kcell.py
```diff
@@ -1664,48 +1757,9 @@ def __copy__(self) -> KCell:
...
-        if self._locked:
-            raise LockedError(self)
-        self.ports.add_ports(
-            ports=ports, prefix=prefix, suffix=suffix, keep_mirror=keep_mirror
-        )
-
     @classmethod
     def from_yaml(
-        cls: Callable[..., KCell],
+        cls,
```
sebastian-goeldi, line 1762: why remove the type hint here?

src/kfactory/kcell.py
```diff
@@ -2569,6 +2608,7 @@ def to_yaml(cls, representer, node):  # type: ignore
             name: _serialize_setting(info)
             for name, info in node.info.model_dump().items()
         }
+        print(cls)
```
sebastian-goeldi, line 2611: forgot to delete this?

src/kfactory/kcell.py
```diff
@@ -5886,6 +5822,11 @@ def dcenter(self) -> kdb.DPoint:
         """Coordinate of the port in um."""
         return self.bbox().center()
 
+    def insert(self, cell: KCell | VKCell) -> VInstance:
+        vi = VInstance(cell)
+        self.insts.append(vi)
+        return vi
```
sebastian-goeldi, line 5828: this would be the same as `insert_vinst`, correct? If so, just delegate to that one instead of duplicating code.

sebastian-goeldi:
Could it be that `VInstance` and `Instance` are still swapped in the code?

src/kfactory/kcell.py
MatthewMckee4: Recursion error as when you call self._kdb_cell, since its a private attr, it calls self.__getattr__("_kdb_cell"), so you need to call super().__getattr__ to get the cell, and if you do self.shapes for example it will fail the super get attr and do the _kdb_cell get attr

src/kfactory/kcell.py
MatthewMckee4: Just seemed weird to type hint cls in this instance, i also added return type of Self, i think this is correct here

MatthewMckee4:
> Could it be that `VInstance` and `Instance` are still swapped in the code?

What are you referring to?

MatthewMckee4:
These failed tests are to do with the hashing function

sebastian-goeldi:
> > Could it be that `VInstance` and `Instance` are still swapped in the code?
> 
> What are you referring to?

I believe the two classes are still out of order between the original version, makes review a bit hard. I will check locally

src/kfactory/kcell.py
sebastian-goeldi: fair enough

The problem with self-type is it's only >3.11

As google colab is dragging their feet (3 years for now I think), we need compatibility with 3.10 for now. But I believe you can import it from `typing_extensions` to make it work

src/kfactory/kcell.py
sebastian-goeldi: I see, then maybe we make a match/ifelse for kdb_cell?

src/kfactory/kcell.py
MatthewMckee4: Can change to import from typing extensions, I have a habit of forgetting this and importing from typing always

MatthewMckee4:
> > > Could it be that `VInstance` and `Instance` are still swapped in the code?
> 
> > 
> 
> > What are you referring to?
> 
> 
> 
> I believe the two classes are still out of order between the original version, makes review a bit hard. I will check locally

I changed the order back, but git still seeing it as many more lines changed unfortunately

sebastian-goeldi:
> > > > Could it be that `VInstance` and `Instance` are still swapped in the code?
> > 
> > 
> > > 
> > 
> > 
> > > What are you referring to?
> > 
> > 
> > I believe the two classes are still out of order between the original version, makes review a bit hard. I will check locally
> 
> I changed the order back, but git still seeing it as many more lines changed unfortunately

I see. I can try to see whether I can make it work locally, no worries.

src/kfactory/kcell.py
sebastian-goeldi: Oh, I see, that might not work. Hmm, then let's keep it like this for now

src/kfactory/kcell.py
MatthewMckee4: Already imported from typing extensions

MatthewMckee4:
I moved some of it back around incorrectly, will fix that

joamatab:
Something is not right here

after merging this gdsfactory hangs
https://github.com/gdsfactory/gdsfactory/actions/runs/12621427365/job/35168279335

MatthewMckee4:
I must've not changed the uv test command back from my PR that added uv

MatthewMckee4:
I don't know why it doesn't error but pytest xdist shouldn't be installed

PR #546, opened 2025-01-12
KCell/DKCell/VKCell - Instance/DInstance/VInstance base classes

sebastian-goeldi:
## Summary by Sourcery

Refactor KCell to inherit from KObject, introducing common bounding box properties and methods.

Enhancements:
- Introduce KObject as a base class for common bounding box operations.
- Added BBoxConstructor, PointLike, and BoxLike protocols for type hinting and interface definition.
- Refactored bounding box related properties in KCell to use the new KObject interface.


MatthewMckee4:
Just want to put this here as first thoughts and so i can refer to this (this will almost definitely change)
```mermaid
classDiagram
    BaseCell <|-- BaseKCell
    BaseKCell <|-- KCell
    BaseKCell <|-- DKCell

   KObject <|-- BaseKCell

    BaseCell <|-- ComponentBase
    BaseCell <|-- VKCell

    ComponentBase <|-- Component
    DKCell <|-- Component

    VKCell <|-- ComponentAllAngle
    ComponentBase <|-- ComponentAllAngle
```

src/kfactory/kcell.py
```diff
@@ -1487,41 +1505,192 @@ def copy(
...
+        return self.ibbox().center().y
+
+    @property
+    def icenter(self) -> tuple[int, int]:
+        """Coordinate of the port in um."""
+        center = self.ibbox().center()
+        return center.x, center.y
+
+
+class BaseKCell(KObject[KUnit]):
```
MatthewMckee4, line 1675: Consider making this inherit from `BaseModel` once everything else works

src/kfactory/kcell.py
```diff
@@ -1680,16 +1854,44 @@ def __init__(
             size_info=SizeInfo(_kdb_cell.bbox),
             dsize_info=DSizeInfo(_kdb_cell.dbbox),
             insts=insts,
+            info=Info(**(info or {})),
+            bbox=_kdb_cell.bbox,
+            dbbox=_kdb_cell.dbbox,
+            ibbox=_kdb_cell.bbox,
```
MatthewMckee4, line 1860: I believe we can remove these as we have defined functions already

src/kfactory/kcell.py
```diff
@@ -6899,21 +7122,21 @@ def layer_info(self) -> kdb.LayerInfo:
...
 
     @property
-    def width(self) -> int:
+    def iwidth(self) -> int:
         """Width of the port. This corresponds to the width of the cross section."""
         return self.cross_section.width
 
     @classmethod
-    def from_yaml(cls: type[Port], constructor, node) -> Port:  # type: ignore
+    def from_yaml(cls: type[BasePort], constructor, node) -> BasePort:  # type: ignore
```
MatthewMckee4, line 7130: Could make this generic, replace BasePort with TBasePort.

src/kfactory/kcell.py
```diff
@@ -6928,7 +7151,7 @@ def copy(
         self,
         trans: kdb.Trans | kdb.DCplxTrans = kdb.Trans.R0,
         post_trans: kdb.Trans | kdb.DCplxTrans | None = None,
-    ) -> Port:
+    ) -> BasePort:
```
MatthewMckee4, line 7154: Could make this return Self or generic over TPort

src/kfactory/kcell.py
```diff
@@ -7302,7 +7544,144 @@ def print(self, type: Literal["dbu", "um", None] = None) -> None:
...
+        mirror flag.
+        """
+        return self.trans.angle
+
+    @angle.setter
+    def angle(self, value: int) -> None:
+        self.iangle = value
+
+    @property
+    def x(self) -> int:
```
MatthewMckee4, line 7563: I think a lot of these should return super().i<function_name> to reduce code duplication

src/kfactory/kcell.py
```diff
@@ -8419,7 +8834,7 @@ def update_default_trans(
 
 
 def pprint_ports(
-    ports: Iterable[Port], unit: Literal["dbu", "um", None] = None
+    ports: Iterable[BasePort | Port | DPort], unit: Literal["dbu", "um", None] = None
```
MatthewMckee4, line 8837: can even just have Iterable[BasePort]

src/kfactory/kcell.py
sebastian-goeldi: no the problem is that it cannot return Self. Self only makes sense if it's the same, object, or am I misinterpreting the type?

And we don't have generics (yet)

src/kfactory/kcell.py
sebastian-goeldi: fair enough

PR #559, opened 2025-01-23
Fix more tests

MatthewMckee4:
## Summary by Sourcery

Refactor the `Port` class to use width instead of dwidth, and update related classes and tests.


sebastian-goeldi (approved):
Thank you ! this is great

PR #565, opened 2025-01-24
add Tunit, i and d properties for cells

MatthewMckee4:
## Summary by Sourcery

Add Tunit, i, and d properties to handle bounding box coordinates and dimensions in different units for cells, enhancing the ProtoKCell class with abstract methods for bounding box calculations.

New Features:
- Introduce Tunit, i, and d properties for cells to handle bounding box coordinates and dimensions in different units.

Enhancements:
- Add abstract methods for bounding box calculations in ProtoKCell class to support different unit types.


src/kfactory/kcell.py
```diff
@@ -1956,6 +1957,15 @@ class BaseKCell(BaseModel, ABC, arbitrary_types_allowed=True):
...
     _base_kcell: BaseKCell
 
+    @abstractmethod
+    def bbox(self, layer: LayerEnum | int | None = None) -> BoxLike[TUnit]: ...
+
+    @abstractmethod
+    def ibbox(self, layer: LayerEnum | int | None = None) -> BoxLike[int]: ...
+
+    @abstractmethod
+    def dbbox(self, layer: LayerEnum | int | None = None) -> BoxLike[float]: ...
```
sebastian-goeldi, line 1967: These don't need to be abstract do they? They can directly be

```suggestion
    def ibbox(self, layer: LayerEnum | int | None = None) -> kdb.Box:
        return self._base_kcell.bbox(layer) # type: ignore[attr] <- forgot the actual name but they work

    def dbbox(self, layer: LayerEnum | int | None = None) -> kdb.DBox:
         return self._base_kcell.dbbox(layer) # type: ignore[attr]
```

src/kfactory/kcell.py
```diff
@@ -3250,96 +3260,142 @@ def get_meta_data(
...
+    @abstractmethod
+    def ymax(self) -> TUnit: ...
+
+    @property
+    @abstractmethod
+    def xsize(self) -> TUnit: ...
+
+    @property
+    @abstractmethod
+    def ysize(self) -> TUnit: ...
```
sebastian-goeldi, line 3317: Do these need to be abstract? we can just return the proper values from `self.bbox()`

src/kfactory/kcell.py
MatthewMckee4: these are in ProtoKCell so here they do need to be abstract

src/kfactory/kcell.py
MatthewMckee4: good point

src/kfactory/kcell.py
MatthewMckee4: they are implemented in ProtoTKCell

sebastian-goeldi: oh, I see. My bad

sebastian-goeldi (approved):

sebastian-goeldi:
Feel free to merge when ready

sebastian-goeldi (approved):
Thx Matthew, looks great!

PR #576, opened 2025-01-26
Fix cell decorator

MatthewMckee4:
## Summary by Sourcery

Update the type signature of the `@cell` decorator to support generic cell types. Remove the `@dcell` decorator.

New Features:
- Support generic cell types in the `@cell` decorator.

Tests:
- Updated type annotations.


MatthewMckee4:
Ah will fix that tomorrow

PR #590, opened 2025-01-29
Add more ruff rules and performed ruff check fix

MatthewMckee4:
## Summary by Sourcery

Add `strict=False` to all zip functions.

Bug Fixes:
- Fixed a potential TypeError when using zip with iterables of different lengths.

Chores:
- Applied ruff check fix.


MatthewMckee4:
This is the wrong approach, ill implement each rule one at a time

PR #615, opened 2025-02-13
change postprocess to get the proper output of the wrapped function

sebastian-goeldi:


sebastian-goeldi:
Not 100% it's the right type, please check that @MatthewMckee4 . Generally, though, I think it makes sense that we allow the user to post-process a user-level object, i.e. a kcell or similar, not a base kcell

MatthewMckee4 (approved):

MatthewMckee4:
Looks good, the test should be easy enough to fix, I can get to it in a bit

PR #629, opened 2025-03-02
make length attribute in ManhattanRoute configurable with length_function

sebastian-goeldi:
Not finished yet, but this will allow creating calculating `ManhattanRoute` length with custom functions

@joamatab


joamatab:
nice, looking forward to it!

PR #634, opened 2025-03-07
by default ignore collisions

joamatab:
## Summary by Sourcery

Make the router ignore collisions by default, adding an 'ignore' option to the `on_collision` parameter.

Enhancements:
- Change the default value of the `on_collision` parameter in the `route_bundle` function to 'ignore'.
- Add an 'ignore' option to the `on_collision` parameter to disable collision checking.


flaport:
@joamatab I did some digging. Kfactory already supports ignoring the route collisions. Just set on_collision=None. Or in yaml:

image

sebastian-goeldi:
Yes, correct, pass None and it won't do anything. And changing the default behavior to ignore is ridiculous. That means locally you just get no warning at all. It just stops?

flaport:
> For yaml this probably makes sense. So I would suggest change the default strategy for yaml. But for a python flow it's better to abort.

**I 100% disagree.**

Always better to do what you can in stead of erroring out when you're making a chip layout. At least in that case you can more easily make changes.

I'm alright with issuing warnings, but the route should still be built. Even if its wrong.

sebastian-goeldi:
You want to put a few 100k or potentially millions of dollars on the line with a warning?

Not all routing collisions are detectable late on. If you are unlucky, they pass DRC and become more or less undetectable unless you spot them with an eye check.

PR #643, opened 2025-03-22
Add more tests

MatthewMckee4:
## Summary by Sourcery

Adds more tests to improve code coverage and reliability, including tests for utilities, DKCell locking, ports in cells, and instance ports. It also introduces a new simplify utility and its corresponding tests.

Tests:
- Adds tests for utilities, DKCell locking, ports in cells, and instance ports.
- Introduces tests for the new simplify utility.


src/kfactory/utilities.py
```diff
@@ -164,7 +164,7 @@ def pprint_ports(
                     t = port.dcplx_trans
                     dx = t.disp.x
                     dy = t.disp.y
-                    dwidth = port.kcl.to_um(port.cross_section.width)
+                    dwidth = port.cross_section.width
```
sebastian-goeldi, line 167: this should stay in `um` ,no?

sebastian-goeldi (approved):

src/kfactory/utilities.py
MatthewMckee4: Just thought we shouldnt do any conversions in the None branch, but could also remove it and set a default for unit?

MatthewMckee4: Okay thats fair, i dont mind either way, ill revert

Issue #437, opened 2024-07-23
Support `kcl.cell(include_module=True)`


95-martin-orion:
**Is your feature request related to a problem? Please describe.**
gdsfactory v7.* supported `include_module` in cells, which allows the same function name to be used for cells in different modules without worrying about collisions. This is distinct from `kcl.cell(basename="foo")`, which cannot automatically include the name of the function in the cell name.

**Describe the solution you'd like**
Add the `include_module` behavior from `gf.cell` in gdsfactory v7.* to `kcl.cell` and provide access to it in gdsfactory.

For example, this should work:
```python
# In module `foo`, with gdsfactory>=8.0
import gdsfactory as gf

@gf.cell(include_module=True)
def my_comp() -> gf.Component:
    ...

my_comp()  # Creates a cell named e.g. `foo_my_comp`
```

sebastian-goeldi:
I think this is just fighting symptoms and actually a bad idea:

Imagine the following:

<details open>

<summary>bend.py</summary>

```python
import kfactory as kf

@kf.kcl.cell(include_module=True)
def euler(radius: kf.kf_types.um, width: kf.kf_types.um, layer: kf.LayerEnum, enclosure: kf.LayerEnclosue) -> kf.KCell:
    c = kf.kcl.kcell()
    # ...
    return c


@kf.kcl.cell
def mycell() -> kf.KCell:
    c = kf.kcl.kcell()
    c << euler(args1)
    # ...
    return c
```

</details>

<details open>

<summary>other.py</summary>

```python

@kf.kcl.cell
def myothercell() -> kf.KCell:
    c = kf.kcl.kcell()
    c << kf.cells.euler.bend_euler(args1)
    # ...
    return c
```

</details>

<details open>

<summary>main.py</summary>

```python
import kfactory as kf
from bend import mycell
from other import myothercell

@kf.kcl.cell
def mymaincell() -> kf.KCell:
    c = kf.kcl.kcell()
    c << mycell()
    c << myothercell()
    # ...
    return c

mymaincell().show()
```

</details>


I would not want to debug this in a project

95-martin-orion:
Our project follows the Google style guide for Python, which [prohibits importing functions directly](https://google.github.io/styleguide/pyguide.html#22-imports). This is useful because it means you don't have to check your entire repo for `mycell()` to see if you can define it in `other.py` - the module can always be used to differentiate it.

[...]

PR #673, opened 2025-04-11
avoid naming collisions

tvt173:
adds the module name into the naming calculus to avoid collisions

## Summary by Sourcery

Enhancements:
- Update cell name generation to incorporate the module name, improving uniqueness of generated cell names


sebastian-goeldi:
This was requested before, and I still think it's a bad idea. https://github.com/gdsfactory/kfactory/issues/437

You can create very confusing errors if you are not careful, apart from having non-trivial name collisions. And as I said in that issue, I'd never want to debug that.

Also, this would make any gdsfactory component nearly unreadable. The name gets cut of even before any parameter is used:

image

joamatab:
No longer needed i believe

PR #675, opened 2025-04-13
pin base classes

sebastian-goeldi:


src/kfactory/pin.py
```diff
@@ -0,0 +1,367 @@
...
+    def allowed_angles(self) -> list[Literal[0, 1, 2, 3]]:
+        """Angle of the transformation.
+
+        In the range of `[0,1,2,3]` which are increments in 90°.
+        """
+        return self._allowed_angles
+
+    @allowed_angles.setter
+    def allowed_angles(self, value: list[Literal[0, 1, 2, 3]]) -> None:
+        self._allowed_angles = value
```
MatthewMckee4, line 222:
```suggestion
    def allowed_angles(self, value: Sequence[Literal[0, 1, 2, 3]]) -> None:
        self._allowed_angles = list(value)
```

src/kfactory/pin.py
```diff
@@ -0,0 +1,367 @@
...
+        self.y = value[1]
+
+    @property
+    def x(self) -> TUnit:
+        """X coordinate of the port."""
+        return self.pos.x
+
+    @x.setter
+    @abstractmethod
+    def x(self, value: TUnit) -> None: ...
```
MatthewMckee4, line 266:
```suggestion
    @x.setter
    def x(self, value: TUnit) -> None:
        self.pos = (value, self.y)
```

src/kfactory/pin.py
```diff
@@ -0,0 +1,367 @@
...
+
+    @property
+    def y(self) -> TUnit:
+        """Y coordinate of the port."""
+        return self.pos.y
+
+    @y.setter
+    @abstractmethod
+    def y(self, value: TUnit) -> None: ...
```
MatthewMckee4, line 275:
```suggestion
    @y.setter
    def y(self, value: TUnit) -> None:
        self.pos = (self.x, value)
```
I think we should be able to do this?

sebastian-goeldi: Can yes, but imo `pos` should be like `trans`/`dcplx_trans` in Port. If you want tuple, use `center`

sebastian-goeldi:
closing in favor of https://github.com/gdsfactory/kfactory/pull/722

PR #683, opened 2025-04-22
Create port protocol

iperezgamiz:
## Summary by Sourcery

Refactor port creation protocols and type hints to improve type safety and flexibility for port creation across different unit types

New Features:
- Create new type protocols for port creation with multiple overloaded method signatures
- Add support for more flexible port creation with various transformation and layer specification options

Enhancements:
- Introduce generic CreatePortFunction protocols to standardize port creation methods
- Add type-specific CreatePortFunctionInt and CreatePortFunctionFloat to support different numeric types
- Improve layer handling to support both integer and LayerEnum inputs


MatthewMckee4:
What i think we should do is have another `ABC` that contains all the overloads and also has an abstract `_create_port` that subclasses implement

sebastian-goeldi:
Closing in favor of this https://github.com/gdsfactory/kfactory/pull/703 . Thank you Iñigo. The problem with the protocols was that if we were to use them, the overloads would get overwritten. I didn't realize that when I asked you to do this, sorry.

Issue #698, opened 2025-05-16
how to run tests?


joamatab:
on. a new server im getting this

<img width="467" alt="Image" src="https://github.com/user-attachments/assets/97ad7c20-9993-46de-a7c3-545a88bc2393" />

sebastian-goeldi:
I don't know. python version?

sebastian-goeldi:
can you try again?

PR #699, opened 2025-05-16
fix python versions in make test

sebastian-goeldi:
Maybe fix https://github.com/gdsfactory/kfactory/issues/698


MatthewMckee4:
nice catch!

PR #701, opened 2025-05-16
Add conf, kcell to kf exports

MatthewMckee4:
## Summary

`ty` also raises some imports as errors as these werent exported


sebastian-goeldi:
can you rebase this and run ruff to fix the order? not sure what went wrong, but should be a quick fix

PR #716, opened 2025-05-28
Add pins inigo

iperezgamiz:

### DIFF
--- a/src/kfactory/pin.py
+++ b/src/kfactory/pin.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Sequence
 from copy import copy
-from typing import TYPE_CHECKING, Any, Generic, Literal
+from typing import TYPE_CHECKING, Any, Generic, Literal, overload
 
 import klayout.db as kdb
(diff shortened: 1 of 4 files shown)


src/kfactory/pin.py
```diff
@@ -70,7 +77,18 @@ def transformed(
...
+        if isinstance(post_trans, kdb.DCplxTrans):
+            post_trans_: kdb.Trans | kdb.ICplxTrans = post_trans.to_itrans(self.kcl.dbu)
+        else:
+            post_trans_ = post_trans
+
+        base.pos = trans_ * self.pos * post_trans_
+        base.allowed_angles = {
+            a + (int(trans.angle()) % 360 // 90) + (int(post_trans.angle()) % 360 // 90)
+            for a in self.allowed_angles
+        }
```
sebastian-goeldi, line 90: this is wrong, I think. if it's a `kdb.Trans` then angle is in [0,1,2,3] and they should just be added and `%4` at the end

iperezgamiz: True, thanks for the correction

sebastian-goeldi:
closing in favor of #722

PR #722, opened 2025-06-06
add pins new version

iperezgamiz:


sebastian-goeldi:
I don't think you need `c` on the Pin. It's just a set of names (which should reference port names, but we can let `@cell` take care of that.

Similarly, there's no need for `transform` and `transformed`.

Also, `copy` should copy the set of names internally, but definitely not the cell :smile:

sebastian-goeldi:
Looks good! Can you add that one check?

PR #738, opened 2025-06-22
add port defitions to cell decorator

sebastian-goeldi:


sebastian-goeldi:
Yes, but can we use klayout nomenclature for bboxes i.e. right/top/left/bottom, or does it have to be the compass directions?

joamatab:
I like right, left more as well

PR #744, opened 2025-07-21
remove git dependency

joamatab:
removes git dependency

@flaport


sebastian-goeldi:
No, please just remove the dependency and make it a separate optional dependency. Don't remove the actual saving default behavior. That will really mess with most setups. The build directory is generally a way cleaner implementation than using cwd.

PR #746, opened 2025-07-27
better schematic cells

sebastian-goeldi:
## Summary

Allows creating a python file from a `kf.Schematic` or `kf.DSchematic`

## Test Plan

tests/test_schematic.py and tests/test_gdsfactory.py


docs/make_schematic_diagram.py
```diff
@@ -0,0 +1,22 @@
+from pathlib import Path
```
MatthewMckee4, line 1: Could we put this in a scripts/generate_mkdocs.py file? And use in the future if we need to generate other files before mkdocs build

PR #748, opened 2025-07-31
Ability to get a slice of ports

abdelq:
## Summary

I want the ability to easily get a slice of ports.
It's useful for example if I want every other port in a `gf.c.array` by using something like `[1::2]` for routing.

## Test Plan

Added tests for `Ports` and `DPorts`

### DIFF
--- a/src/kfactory/ports.py
+++ b/src/kfactory/ports.py
@@ -696,11 +696,10 @@ class Ports(ProtoPorts[int], ICreatePort):
         """Get ports by slice."""
 
     def __getitem__(self, key: slice | int | str | None) -> Self | Port:
-        if isinstance(key, slice):
-            return self.__class__(bases=self._bases[key], kcl=self.kcl)
-
         if isinstance(key, int):
             return Port(base=self._bases[key])
+        if isinstance(key, slice):
+            return self.__class__(bases=self._bases[key], kcl=self.kcl)
         try:
             return Port(base=next(filter(lambda base: base.name == key, self._bases)))
         except StopIteration as e:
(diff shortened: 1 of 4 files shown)


abdelq:
I've shared an example of a use case in the description. I'm doing the equivalent in with a list comprehension and that's fine, but could be better. I actually tried to use a slice first and was surprised it wasn't working.

src/kfactory/ports.py
```diff
@@ -680,8 +687,18 @@ def get_all_named(self) -> Mapping[str, Port]:
...
+    @overload
+    def __getitem__(self, key: slice) -> Self:
+        """Get ports by slice."""
+
+    def __getitem__(self, key: slice | int | str | None) -> Self | Port:
+        if isinstance(key, slice):
+            return self.__class__(bases=self._bases[key], kcl=self.kcl)
+
         if isinstance(key, int):
             return Port(base=self._bases[key])
```
sebastian-goeldi, line 703:
```suggestion
        if isinstance(key, int):
            return Port(base=self._bases[key])
        if isinstance(key, slice):
            return self.__class__(bases=self._bases[key], kcl=self.kcl)
```

src/kfactory/ports.py
```diff
@@ -792,8 +809,18 @@ def get_all_named(self) -> Mapping[str, DPort]:
...
+    @overload
+    def __getitem__(self, key: slice) -> Self:
+        """Get ports by slice."""
+
+    def __getitem__(self, key: slice | int | str | None) -> Self | DPort:
+        if isinstance(key, slice):
+            return self.__class__(bases=self._bases[key], kcl=self.kcl)
+
         if isinstance(key, int):
             return DPort(base=self._bases[key])
```
sebastian-goeldi, line 825:
```suggestion
        if isinstance(key, int):
            return DPort(base=self._bases[key])
        if isinstance(key, slice):
            return self.__class__(bases=self._bases[key], kcl=self.kcl)
```

sebastian-goeldi (changes requested):

PR #764, opened 2025-08-16
add mirror to SchemaInstance connect

iperezgamiz:
## Summary

Allow mirrored connections between SchemaInstances 
Change Connection from RootModel to BaseModel including a mirror flag

## Pending

Update Schema model validation from YAML to account for mirror flag in Connection
Update Connection.from_list()
Add test

### DIFF
--- a/src/kfactory/layer.py
+++ b/src/kfactory/layer.py
@@ -64,6 +64,9 @@ class LayerInfos(BaseModel):
                 )
         return self
 
+    def __getitem__(self, value: str) -> kdb.LayerInfo:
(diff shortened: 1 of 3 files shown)


sebastian-goeldi:
Oh, hold on. Now I remember why this does not work. A connection is not unidirectional anymore it's bidirectional, therefore with the Connection(..., mirror=True) you cannot fix this issue (it only works on paper if the the second element stays the second element in the connection, which it likely does not)

That's why I can up with the `MirrorPlacement` class, to indicate that a component should be flipped

sebastian-goeldi (approved):

PR #772, opened 2025-08-26
Remove require label

MatthewMckee4:
## Summary

Frankly I think it is often annoying to have this test fail.

I believe we should be add labels when we need, and sometimes the label we want to add isn't one of the ones it requires.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -21,17 +21,3 @@ jobs:
       - uses: release-drafter/release-drafter@v6
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  require_label:
-    if: github.event.pull_request
-    needs: update_release_draft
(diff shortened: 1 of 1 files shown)


MatthewMckee4:
@sebastian-goeldi I am not super fussed about this, i just dont always like the failing test shouting at me. And we should be more intentional in the labels we add.

sebastian-goeldi:
Unfortunately github has no really "allowed to fail". So I don't know about this. We can also rely more on drafting the release ourselves. It's a bit annoying for automatic changelog then though :/

sebastian-goeldi:
Would a compromise be to not have it fail if it's a draft but has no labels yet?
Or if we can somehow to get it to block unless it has a label, I'd be ok with that too

sebastian-goeldi:
I'll close fore now. I think one option would be to just have some labels which will not affect changelog (e.g. the github_action one does)

PR #783, opened 2025-09-05
Instead of raising a ValueError when a cell with the same name already exists, it now automatically creates a new cell with a suffix (e.g., $1, $2) and issues a warning.

joamatab:
## Summary by Sourcery

Change create_cell to auto-rename cells with duplicate names by appending a $n suffix and warn instead of raising ValueError

Enhancements:
- Automatically append a unique suffix to duplicate cell names and create the new cell instead of raising an error
- Issue a warning when a cell name collision triggers automatic renaming

### DIFF
--- a/src/kfactory/layout.py
+++ b/src/kfactory/layout.py
@@ -1310,12 +1310,23 @@ class KCLayout(
 
         """
         with self.thread_lock:
-            if allow_duplicate or (self.layout_cell(name) is None):
+            if allow_duplicate:
                 return self.layout.create_cell(name, *args)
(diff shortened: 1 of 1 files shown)


MatthewMckee4:
I am okay with this if @sebastian-goeldi is. Also can we remove sourcery on this repo please?

sebastian-goeldi:
Already supported although slightly hidden:

```python
    c = gf.Component(
        kdb_cell=gf.kcl.create_cell(
            duplicate_cell_name, allow_duplicate=True
        )
    )
```

PR #796, opened 2025-09-19
skip electrical orphan

joamatab:


joamatab:
this does not detect opens

PR #824, opened 2025-10-29
Fix inconsistent angle assignment via `Port.angle` instead of `Port.trans.angle`

### LINKED ISSUE #823: Inconsistent behavior when assigning trans.angle vs. angle on a Port
In kfactory/routing/optical.py line 798 in `place_manhattan()` assigning to `route_start_port.trans.angle` sometimes fails silently, the angle value remains unchanged. I've notice that assigning to `route_start_port.angle` instead works consistently in my testing.

Assigning to `route_start_port.trans.angle` results in inconsistent rotation behavior for virtual ports. When the routing is between ports that exist on a different component it seems to work just fine.

```
import gdsfactory as gf
c = gf.Component()
pad = c << gf.c.pad(size=(15,15))
pad.move((100, 0))

port1 = c.add_port(
    "e1",
    center=(10, 0),
    width=15,
    orientation=0,
    layer='M3',
    port_type='electrical'
)
c.draw_ports()

route = gf.routing.route_single(
    c,
    port1=port1,
    port2=pad.ports['e1'],
    layer='M3',
    on_error=None,
    bend="wire_corner",
    route_width=15,
)

# same issue when using a different router
route = gf.routing.route_bundle_electrical(
    c,
    ports1=[port1],
    ports2=[pad.ports['e1']],
    layer='M3',
    route_width=15,
)

c
```

This code should place a metal trace between and connecting the virtual port (port1) and the pad port ("e1"), but it incorrectly rotates the straight object on the first port.

Image

When routing between ports of two separate pads references the rotation of the port in `place_manhattan()` works just fine.

```
import gdsfactory as gf

c = gf.Component()

pad = c << gf.c.pad(size=(15,15))
pad.move((0, 0))
pad2 = c << gf.c.pad(size=(15,15))
pad2.move((100, 0))

route = gf.routing.route_bundle_electrical(
    c,
    ports1=[pad.ports['e3']],
    ports2=[pad2.ports['e1']],
    layer='M3',
    route_width=15,
)
c
```

Image

cjcarver:
Assigning to Port.trans.angle sometimes fails silently when using a virtual port.
This PR updates `place_manhattan()` to modify Port.angle instead.

Fixes #823

### DIFF
--- a/src/kfactory/ports.py
+++ b/src/kfactory/ports.py
@@ -631,7 +631,10 @@ class DCreatePort(ABC):
                     kcl=self.kcl,
                 )
             else:
-                trans_ = kdb.Trans(rot=int(dcplx_trans.angle // 90), u=self.kcl.to_dbu(dcplx_trans.disp))
+                trans_ = kdb.Trans(
+                    rot=int(dcplx_trans.angle // 90),
+                    u=self.kcl.to_dbu(dcplx_trans.disp),
+                )
                 port = DPort(
                     name=name,
                     trans=trans_,
--- a/src/kfactory/routing/optical.py
+++ b/src/kfactory/routing/optical.py
@@ -793,7 +793,7 @@ def place_manhattan(
             "place_manhattan needs to be passed a fixed bend90 cell with two optical"
             " ports which are 90° apart from each other with port_type 'port_type'."
         )
-    
+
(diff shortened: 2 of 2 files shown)


sebastian-goeldi:
Hi, sorry, maybe I should have worded it a bit differently. I think #825 addresses this properly and will give a warning instead of an error (as in most cases we can reasonably recover from this problem)

PR #832, opened 2025-11-28
Don't set anchor if it's not set

abdelq:
## Summary

If `anchor` isn't set, it defaults to `FixedAnchor(x=None, y=None)` when it should be `None`.

## Test Plan

Manually on a real schematic


sebastian-goeldi:
Thanks!

PR #882, opened 2026-02-09
Fix insert_vinsts and use its recursive argument

abdelq:
## Summary

The intersection is done at the wrong spot leading to empty set. Also `recursive` was not being used.

## Test Plan

Tested on private repo's code.


dansgithubuser:
I'm curious what's the point of the `& self.kcl.tkcells.keys()`. Is it ever not a no-op?

sebastian-goeldi:
> I'm curious what's the point of the `& self.kcl.tkcells.keys()`. Is it ever not a no-op?

<del>In case it's been manually manipulated yes. This can theoretically happen if someone changes `kcl` manually. It can crash really hard if that happens, and the klayout errors can be very no fun when that happens</del>

Sorry, maybe I misremember. But I believe if you clean up the layout object, it can have an impact, but I can't remember exactly why I added it. But it was related to having klayout segfault bc accessing an invalid index at some point

PR #957, opened 2026-05-11
fix: use original BasePort ids for pin port mapping in dup()

flaport:
## Summary

- `dup()` crashes with `KeyError` when copying a cell that has pins
- `Ports.__iter__` yields fresh `Port(base=b)` wrappers each call, so `id(p)` from `enumerate(c.ports)` produces ephemeral ids that never match the `BasePort` objects stored in `pin.ports`
- Fix: build `port_mapping` from `self.ports._bases` — the same objects that `create_pin` captured via `port.base`

## Test plan

- [ ] Copy a component with pins: `c.dup()` / `c.copy()` should preserve pins without `KeyError`
- [ ] Verify the duplicated cell's pin ports reference the new cell's ports, not the original's

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sebastian-goeldi (approved):

sebastian-goeldi:
I think this has a problem though. The ports are now shared between the dup and non-dup

sebastian-goeldi:
I fixed this in a different branch as it needed to be backported and not target main

PR #969, opened 2026-05-20
Avoid polygon changes from a VCell -> Cell conversion

abdelq:
## Summary

Avoid implicit polygon changes from a VCell -> Cell conversion.
This was partially handled with the `to_itype` change, but `merge` is doing this too.
Since it's happening on the parent's shape and the rest and child instances, it's a better default to avoid the merge.

There is some cleanup done as separate commits. Some dead code likely from refactors.

### DIFF
--- a/src/kfactory/instance.py
+++ b/src/kfactory/instance.py
@@ -797,14 +797,6 @@ class VInstance(ProtoInstance[float], UMGeometricObject):
                 cell_.name = cell_name
                 for port in self.cell.ports:
                     cell_.add_port(port=port.copy(trans_))
-                for c_shapes in (
-                    cell_.shapes(layer) for layer in cell_.kcl.layer_indexes()
-                ):
-                    if not c_shapes.is_empty():
-                        r = kdb.Region(c_shapes)
-                        r.merge()
-                        c_shapes.clear()
-                        c_shapes.insert(r)
                 settings = self.cell.settings.model_copy()
                 settings_units = self.cell.settings_units.model_copy()
                 cell_.settings = settings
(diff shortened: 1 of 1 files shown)


src/kfactory/instance.py
```diff
@@ -927,17 +917,10 @@ def insert_into_flat(
...
-            else:
-                for layer, shapes in self.cell.shapes.items():
-                    for shape in shapes.transform(trans * self.trans):
-                        cell.shapes(layer).insert(shape)
-                for vinst in self.cell.vinsts:
-                    vinst.insert_into_flat(cell, trans=trans * self.trans)
+            for layer in self.cell.kcl.layer_indexes():
+                reg = kdb.Region(self.cell.kdb_cell.begin_shapes_rec(layer))
+                reg.transform(kdb.ICplxTrans((trans * self.trans), cell.kcl.dbu))
+                cell.shapes(layer).insert(reg)
```
sebastian-goeldi: I think here there might be a logic error. It should check whether cell is a ProtoTKCell and not `self.cell`

(this was already false in the old code, which I thought I fixed at some point)

sebastian-goeldi:
I think though you must merge when doing this. Because if you don't, rotation does not guarantee keeping coincidence unless points match. You can only skip this if the resulting transformation falls on grid and is keeping a manhattan orientation

src/kfactory/instance.py
abdelq: Moving this stuff to https://github.com/gdsfactory/kfactory/pull/973

abdelq:
Closing, until I think of something better.

PR #997, opened 2026-06-26
Detect duplicate cell names before writing layouts

joamatab:
## Summary
- Add `DuplicateCellNameError` and a pre-write check (`_check_duplicate_cell_names`) that validates all cells in the write set have unique names before writing GDS/OASIS files
- The error message details which names are duplicated, how many cells share each name, and suggests enabling `kf.config.debug_names` for earlier detection
- Fix pre-existing `ty` type-checker false positive for `os.sched_getaffinity` on macOS

## Test plan
- [ ] Verify writing a layout with unique cell names still works normally
- [ ] Verify writing a layout with duplicate cell names raises `DuplicateCellNameError` with a clear message
- [ ] Run existing test suite to check for regressions

### DIFF
--- a/src/kfactory/conf.py
+++ b/src/kfactory/conf.py
@@ -181,14 +181,15 @@ def get_affinity() -> int:
     On (most) linux we can get it through the scheduling affinity. Otherwise,
     fall back to the multiprocessing cpu count.
     """
-    if hasattr(os, "sched_getaffinity"):
-        return len(os.sched_getaffinity(0))
     try:
-        import multiprocessing
+        return len(os.sched_getaffinity(0))
+    except AttributeError:
+        try:
+            import multiprocessing
 
-        return multiprocessing.cpu_count()
-    except ModuleNotFoundError:
-        return 1
+            return multiprocessing.cpu_count()
+        except ModuleNotFoundError:
+            return 1
 
 
 dotenv_path = find_dotenv(usecwd=True)
--- a/src/kfactory/conf.py
+++ b/src/kfactory/conf.py
@@ -181,15 +181,14 @@ def get_affinity() -> int:
     On (most) linux we can get it through the scheduling affinity. Otherwise,
     fall back to the multiprocessing cpu count.
     """
+    if hasattr(os, "sched_getaffinity"):
+        return len(os.sched_getaffinity(0))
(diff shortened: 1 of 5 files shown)


src/kfactory/conf.py
```diff
@@ -181,15 +181,14 @@ def get_affinity() -> int:
...
-            import multiprocessing
+    try:
+        import multiprocessing
 
-            return multiprocessing.cpu_count()
-        except ModuleNotFoundError:
-            return 1
+        return multiprocessing.cpu_count()
+    except ModuleNotFoundError:
+        return 1
```
sebastian-goeldi: Why are you changing this behavior? This has nothing to do with the PR. Also, afaik on platforms which do not support the scheduler affinity, this will error out

src/kfactory/kcell.py
```diff
@@ -482,125 +527,59 @@ def name(self, value: str) -> None:
...
-                        "Name conflict in "
-                        f"{frame_info.filename}"
-                        f"{function_name} at line {frame_info.lineno}\n"
-                        f"Renaming {self.name} (cell_index="
-                        f"{self.kdb_cell.cell_index()}) to"
-                        f" {value} would cause it to be named the same as:\n"
-                        + "\n".join(
-                            f" - {tkcell.name} "
-                            f"(cell_index={tkcell.kdb_cell.cell_index()}),"
-                            f" function_name={tkcell.function_name},"
```
sebastian-goeldi: so, you just drop the entire logic to figure out where duplicate names come from? cool, but no

src/kfactory/kcell.py
```diff
@@ -1367,6 +1349,9 @@ def write_bytes(
             case _:
                 ...
 
+        relevant_cells = {self.cell_index(), *self.called_cells()}
+        _deduplicate_cell_names(self.layout(), relevant_cells)
+
```
sebastian-goeldi, line 1506: silently renaming cells, no. At least add an option and make it false by default

sebastian-goeldi (changes requested):

discdiver:
@joamatab indicated high priority for a user

sebastian-goeldi (approved):

PR #1001, opened 2026-06-30
ignore ty warning on affinity

KRRT7:
## Summary
- avoid calling os.sched_getaffinity on platforms that do not provide it
- fall back to multiprocessing.cpu_count() for ty worker count detection

## Verification
- uv run ty check src/kfactory
- uv run pytest -q

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,6 +167,7 @@ skip = "pyproject.toml, uv.lock"
 [tool.ty]
 src.exclude = [ "docs", "src/kfactory/widgets/interactive.py", "tests/gdsfactory-yaml-pics" ]
 src.include = [ "docs", "src", "tests" ]
+environment.python-platform = "all"
 
 [tool.pytest]
 ini_options.testpaths = [ "src", "tests" ]
--- a/src/kfactory/conf.py
+++ b/src/kfactory/conf.py
@@ -181,15 +181,15 @@ def get_affinity() -> int:
     On (most) linux we can get it through the scheduling affinity. Otherwise,
     fall back to the multiprocessing cpu count.
     """
+    sched_getaffinity = getattr(os, "sched_getaffinity", None)
+    if sched_getaffinity is not None:
+        return len(sched_getaffinity(0))
     try:
-        return len(os.sched_getaffinity(0))
-    except AttributeError:
-        try:
-            import multiprocessing
+        import multiprocessing
 
-            return multiprocessing.cpu_count()
-        except ModuleNotFoundError:
-            return 1
+        return multiprocessing.cpu_count()
+    except ModuleNotFoundError:
+        return 1
 
 
 dotenv_path = find_dotenv(usecwd=True)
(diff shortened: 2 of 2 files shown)


sebastian-goeldi:
not reviewing pure LLM PRs

KRRT7:
@sebastian-goeldi Hi, I had let @joamatab  know I'd be opening a few PRs, mainly perf related.

sebastian-goeldi:
Then I would ask you to not do pure LLM PRs. If you make changes, you can fully explain why you make certain changes. I don't want to just get PRs which do not explain what they are doing and explanations on why it's better. I can ask Claude myself to do a performance analysis and then at least I have the context for it.

But this for example is a pure stylistic choice and a guess on what your most likely platform is to run on. Functionally the runtime difference will likely not be measurable because this is a once run at import time.

KRRT7:
it's not a stylistic choice, os.sched_getaffinity is not available on macOS.

sebastian-goeldi:
and the try/catch in the original code is for decoration?

if that were true all the macos runners would fail

and again, the difference is minuscule and depends on what your expected result is, found or not found
https://stackoverflow.com/a/903238

KRRT7:
```
uv run ty check src/kfactory
error[unresolved-attribute]: Module `os` has no member `sched_getaffinity`
   --> src/kfactory/conf.py:185:20
    |
185 |         return len(os.sched_getaffinity(0))
    |                    ^^^^^^^^^^^^^^^^^^^^
    |
info: The member may be available on other Python versions or platforms
info: Python 3.12 was assumed when resolving the `sched_getaffinity` attribute
  --> pyproject.toml:11:19
   |
11 | requires-python = ">=3.12"
   |                   ^^^^^^^^ Python version configuration
   |

Found 1 diagnostic
```

this was my original motivation behind the PR.

sebastian-goeldi:
Yes, hence stylistic choice. And that was exactly my point about the PRs.

You did not include this in the initial PR description. And then I could have figured out that it's from the type checker. Which fair is probably worth fixing, but the fix in this case is not getattr, but rather to tell ty to ignore this line, or describe in the PR why this is preferable over the existing implementation. Context is important!

KRRT7:
@sebastian-goeldi apologies, I misunderstood what you were saying and yeah, sorry about that, will update the bodies and provide context, the other PRs were manually reviewed by me before I opened, so I'm confident about them.

KRRT7:
I'll have ty ignore the warning here then

sebastian-goeldi:
Apologies. If you review the PRs manually then it's all fine. I initially thought they are pure Codex PRs, in which case it won't be great. I looked at the other PRs as well more in-depth

KRRT7:
No, I use codex and claude interchangeably, I tend to open multiple PRs at a time since I prefer stacked PRs, but GitHub doesn't support them too well so I have to split them up, but I forget people dont like that

sebastian-goeldi (approved):
ty!

PR #1002, opened 2026-06-30
Add performance hotspot benchmarks

KRRT7:
## Summary
- add a route bundle collision benchmark for checking collision-validation overhead
- add a hotspot benchmark covering named port lookup, extrusion, cell cache, and schematic pin materialization

## Verification
- uv run python benchmarks/route_bundle_collision.py --repeat 5
- uv run python benchmarks/hotspots.py --repeat 5
- pre-commit hooks during commit

### DIFF
--- a/benchmarks/__init__.py
+++ b/benchmarks/__init__.py
@@ -1,0 +1,1 @@
+"""Local benchmark scripts for kfactory development."""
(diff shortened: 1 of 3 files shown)


KRRT7:
plan to reopen with a proper benchmarking suite

PR #1003, opened 2026-06-30
Speed up schematic materialization lookups

KRRT7:
## Summary
- reuse named top-level port maps during schematic materialization
- add an identity fast path for validating pin ports already on the cell
- reuse the materialized instance map instead of repeatedly scanning cell instances

## Benchmark
- schematic_pin_materialize improved from about 124ms median to about 100ms median in benchmarks/hotspots.py on local runs

## Verification
- uv run ruff check src/kfactory/schematic.py src/kfactory/kcell.py
- uv run pytest tests/test_schematic.py tests/test_pins.py tests/test_pins_extra.py -q


sebastian-goeldi:
not reviewing pure LLM PRs

src/kfactory/kcell.py
```diff
@@ -2539,7 +2398,10 @@ def _check_pin_ports_in_cell(
     cell: ProtoTKCell[Any], ports: Iterable[ProtoPort[Any]], *, pin_name: str
 ) -> None:
     cell_ports = cell.base.ports
+    cell_port_ids = {id(port) for port in cell_ports}
     for port in ports:
+        if id(port.base) in cell_port_ids:
+            continue
```
sebastian-goeldi: this seems to not do much though? this seems counter intuitive. I get what it's trying to do, but then it can be instead used in the second check instead of the check there. then it's fine imo

src/kfactory/schematic.py
```diff
@@ -1144,6 +1145,11 @@ def place_port(
         """
         port = self.ports[name]
         if isinstance(port, PortArrayRef):
+            if instances is not None:
```
sebastian-goeldi, line 1149: as far as I can see instances is never None in the actual call, so for effective use the second branch is just dead and they could just be combined. Instance lookup with existing instances seems reasonable

sebastian-goeldi (changes requested):

KRRT7:
@sebastian-goeldi would you consider implementing https://asv.readthedocs.io/en/latest/ style benchmarks on paths you care about?

sebastian-goeldi:
hotpaths are very subjective and highly depend on the specific tapeout. So it's very hard to do them. Additionally, atm I don't really have time to purely focus on performance, I have some hunches but they're just that. I think some workloads I can think of worth investigating:

- cell with high number of instances (e.g. 5000 instances of straights of the same length in one cell)
- high number of cells (e.g. 2000 straights with different lengths)
- high number of trivial polygons (e.g. 2000 boxes of same size)
- low number of highly complex polygons (e.g. 50 polygons with >10k points each; note, these will be split upon save by klayout)

I don't currently have the bandwidth to implement these though.

KRRT7:
> hotpaths are very subjective and highly depend on the specific tapeout. So it's very hard to do them. Additionally, atm I don't really have time to purely focus on performance, I have some hunches but they're just that. I think some workloads I can think of worth investigating:
> 
> * cell with high number of instances (e.g. 5000 instances of straights of the same length in one cell)
> * high number of cells (e.g. 2000 straights with different lengths)
> * high number of trivial polygons (e.g. 2000 boxes of same size)
> * low number of highly complex polygons (e.g. 50 polygons with >10k points each; note, these will be split upon save by klayout)
> 
> I don't currently have the bandwidth to implement these though.

no worries! I can work with that and create the benchmarks.

PR #1004, opened 2026-06-30
Speed up path extrusion point generation

KRRT7:
## Summary
- replace per-point DCplxTrans construction and NumPy angle conversion with direct normal-vector offset math
- preserve existing default end-angle behavior and zero-length segment handling

## Benchmark
- extrude_path improved from about 99ms to about 52ms median in benchmarks/hotspots.py local runs
- extrude_path_dynamic improved from about 151ms to about 63ms median in benchmarks/hotspots.py local runs

## Verification
- uv run ruff check src/kfactory/enclosure.py
- uv run pytest tests/test_extrude.py tests/test_enclosure.py tests/test_factories_utils.py -q

### DIFF
--- a/src/kfactory/enclosure.py
+++ b/src/kfactory/enclosure.py
@@ -8,6 +8,7 @@ region.
 from __future__ import annotations
 
 import itertools
+import math
 import sys
 from collections import defaultdict
 from enum import IntEnum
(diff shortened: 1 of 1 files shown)


sebastian-goeldi:
not reviewing pure LLM PRs

sebastian-goeldi:
what's the speedup here? this is usually not in a very critical path, so small differences in runtime are probably not worth the headache if something's wrong

PR #1005, opened 2026-06-30
Cache named port lookups

KRRT7:
## Summary
- cache named port lookup results for Ports and DPorts
- share the cache across fresh KCell/DKCell/VKCell port wrappers
- validate cached hits against the current BasePort.name so renamed ports keep working
- add regression coverage for renamed ports and duplicate-name first-match behavior

## Benchmark
From the combined local benchmark branch:
- ports[name] improved from about 224ms median to about 38ms median in benchmarks/hotspots.py

## Verification
- uv run ruff check src/kfactory/ports.py src/kfactory/kcell.py tests/test_ports.py
- uv run ty check src/kfactory/ports.py src/kfactory/kcell.py tests/test_ports.py
- uv run pytest tests/test_ports.py tests/test_rename.py tests/test_cell.py tests/test_instance_ports.py -q


sebastian-goeldi:
not reviewing pure LLM PRs

sebastian-goeldi:
I don't think this is possible. The names of ports can change, this only really works once the cell is locked.

And there is no path port -> cell, so you cannot notify the cell about a port name change as the port does not know whether it belongs to a cell (or theoretically two, though that is undesired)

KRRT7:
Good point. I changed this so the name cache is only used for locked cells.

Unlocked cells now keep the original ordered scan behavior, so mutable port names and duplicate-name first-match semantics stay correct. I also added regressions for the stale-cache case.

PR #1011, opened 2026-07-08
Add cell metadata provider API

flaport:
## Motivation

External tools like gfp/Mosaic and simulation runners need structured metadata about cell factories — device type, schematic ports, simulation model links, tags, and display hints. Today the only way to attach this metadata is through `@cell(schematic_function=...)`, which forces PDKs to return a full `DSchematic` object and bury the actual contract inside free-form `s.info` keys (`info["symbol"]`, `info["ports"]`, `info["models"]`). This has several problems:

- **One hook does two jobs.** `schematic_function` is used both for real schematic-driven layout (`@schematic_cell`) and for lightweight symbol/model metadata. Both make `schematic_driven()` return `True`, which is misleading.
- **The real schema is hidden.** The fields that consumers actually depend on (`symbol`, `ports`, `models`) are untyped dict entries in `DSchematic.info`. kfactory doesn't validate them — every PDK and consumer has to duplicate the schema and write its own tests.
- **Port metadata is duplicated.** Each port exists as both an `info["ports"]` dict and a `DSchematic.ports` entry with placeholder cross-sections. These can diverge silently.
- **Consumers depend on private internals.** gfp iterates `kcl.factories._all` and reads `_f_schematic()` because the public API doesn't expose stable metadata access or fully-qualified factory lookup.
- **vcells are excluded.** `@vcell` doesn't accept `schematic_function`, so metadata coverage is incomplete.
- **No composition.** Model metadata, device types, and port descriptions must all live on the original cell decorator. A simulation package can't add models to a geometry package's components without modifying the original source.
- **Schematic function only settable at decoration time.** `schematic_function` can only be passed as a keyword to `@cell(...)`, so it can't be attached after the fact from a different module.

## What this PR adds

A first-class, provider-based cell metadata system that separates structured metadata from schematic-driven layout, plus a `schematic_for` decorator. The metadata types are intentionally minimal — `ports` are `TypedDict` dicts, `models` and `display` are plain `list` and `dict` — so downstream consumers (gfp, PDKs) can define their own schemas without kfactory imposing structure.

### New module: `kfactory.cell_metadata`

| Type | Purpose |
|------|---------|
| `CellMetadata` | Resolved metadata: `device_type`, `ports`, `models`, `tags`, `display`, `info` |
| `PortSpec` | `TypedDict` with required `name`/`kind` and optional `side`/`orientation`/`order`/`cross_section`/`role`/`aliases` |

All metadata is plain dicts and lists at runtime:
- `CellMetadata.ports` — `tuple[PortSpec, ...]` (TypedDict, so plain dicts)
- `CellMetadata.models` — `list` (any structure: dicts, dataclasses, etc.)
- `CellMetadata.display` — `dict`
- `CellMetadata.info` — `dict[str, JSONSerializable]` (free-form, matching existing kfactory/gdsfactory `info` conventions)

kfactory does not validate or interpret model/display contents; that responsibility belongs to the consumer.

### Registration API on `KCLayout`

Seven metadata provider decorators, all following the same `kcl.<field>_for(target)` pattern:

```python
kcl.metadata_for(target)      # full metadata patch
kcl.device_type_for(target)   # device type only
kcl.ports_for(target)         # ports only
kcl.model_for(target)         # model specs (supports position="prepend")
kcl.tags_for(target)          # tags
kcl.display_for(target)       # display hint dict
kcl.info_for(target)          # free-form info dict
```

Plus a schematic decorator:

```python
kcl.schematic_for(target)     # set schematic_function after cell registration
```

Targets can be:
- A **registered factory name** (`"mmi1x2"`) for PDK-local use
- A **fully-qualified name** (`"gdsfactory.components.mmis.mmi1x2.mmi1x2"`) for cross-package overlays or disambiguation
- A **factory object** reference

Metadata providers are plain functions that accept any subset of the cell factory's parameters, enabling parametric metadata (e.g. port count that varies with a cell parameter).

### Resolution API on factory wrappers

Both `WrappedKCellFunc` and `WrappedVKCellFunc` gain:

```python
factory.has_metadata() -> bool
factory.get_metadata(*args, **kwargs) -> CellMetadata
factory.metadata_providers() -> tuple[...]
factory.qualified_name -> str       # stable FQN for external indexing
```

`get_metadata()` parses parameters the same way cell creation does, then resolves all registered providers into a single `CellMetadata`. This is fully independent of `schematic_function` / `get_schematic()`.

### Public helpers on `Factories`

```python
kcl.factories.all()                          # replaces private _all
kcl.factories.annotated()                    # only factories with metadata
kcl.factories.get_by_qualified_name(fqn)     # stable FQN lookup
kcl.factories.get_all_by_name(name)          # all wrappers with a given short name
```

### Merge semantics

When multiple metadata providers target the same factory:
- `device_type`: first value wins; conflicts error unless `replace=True`
- `ports`: replace as a group (port count can be parametric); conflicts error unless `replace=True`
- `models`: list concatenation, with `position="prepend"` support
- `tags`: deterministic set union
- `display`: first value wins; conflicts error unless `replace=True`
- `info`: shallow merge; key conflicts error

## What this PR does NOT change

- No changes to `@schematic_cell`, `TSchematic`, or `DSchematic`
- No `metadata=` / `metadata_function=` kwargs on `@cell` / `@vcell` decorators
- No migration of existing PDK `DSchematic.info` metadata
- No gdsfactory changes

## Example usage

```python
@kcl.cell
def mmi1x2(n: int = 2) -> kf.KCell:
    ...

@kcl.device_type_for("mmi1x2")
def _dt() -> str:
    return "mmi-1x2"

@kcl.ports_for("mmi1x2")
def _ports(n: int = 2) -> list[kf.PortSpec]:
    return [
        kf.PortSpec(name="o1", kind="optical", side="left"),
        *[kf.PortSpec(name=f"o{i+2}", kind="optical", side="right") for i in range(n)],
    ]

@kcl.model_for("mmi1x2")
def _model() -> dict:
    return {"language": "sax", "name": "mmi", "module": "my_models", "qualname": "mmi"}

@kcl.schematic_for("mmi1x2")
def _schematic() -> kf.DSchematic:
    s = kf.DSchematic()
    # ... build schematic ...
    return s

# Resolve metadata for specific parameters:
metadata = kcl.factories["mmi1x2"].get_metadata(n=3)
assert metadata.device_type == "mmi-1x2"
assert len(metadata.ports) == 4
assert metadata.ports[0]["name"] == "o1"
assert metadata.models[0]["module"] == "my_models"
```

## Test plan

- [x] Layout metadata decorators: device_type, ports, models register and resolve correctly
- [x] Parametric providers (port count varies with cell parameter)
- [x] FQN targeting and `position="prepend"` model composition
- [x] Ambiguous/missing target errors
- [x] Provider parameter subset matching and missing-parameter errors
- [x] vcell metadata support
- [x] `schematic_function` and metadata system are independent
- [x] `schematic_for` decorator by name, FQN, and object reference
- [x] Display as plain dict
- [x] Models as plain list (single dict and multi-entry list)

Generated with [Claude Code](https://claude.com/claude-code)


src/kfactory/factory_metadata.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+        )
+        return provider
+
+    def providers_for(
+        self,
+        *,
+        name: str,
+        qualified_name: str,
+        obj: object,
```
sebastian-goeldi, line 133: Are you sure you want to simultaneously filter for name and qname?

sebastian-goeldi: Is object there just model, or what would you like this for exactly?

src/kfactory/factory_metadata.py
```diff
@@ -0,0 +1,328 @@
...
+    provider_kwargs: dict[str, Any] = {}
+    missing: list[str] = []
+    for name, param in sig.parameters.items():
+        if param.kind in (
+            inspect.Parameter.POSITIONAL_ONLY,
+            inspect.Parameter.VAR_POSITIONAL,
+        ):
+            raise TypeError(
+                f"Annotation provider {provider!r} has unsupported parameter {name!r}."
+            )
```
sebastian-goeldi, line 166: Is this to not allow positional args?

You can allow positional ones if they can be mapped to an existing arg though? Can you not?

sebastian-goeldi (approved):

sebastian-goeldi (approved):

PR #1021, opened 2026-07-16
Restore `kcl._future_cell_name` to avoid stale values

abdelq:
## Summary

Context: doing some routing and the routing logic uses `_future_cell_name`
https://github.com/gdsfactory/kfactory/blob/4a70052f42a08d7f110c0a96eddfbf096b45ac56/src/kfactory/routing/generic.py#L149

Having a stale value ends up making failed routes collide on the same cell name (in my case this was due to some exception for some min radius violation in a bend).

I'm restoring the name with a `try/finally` so that it's reset on exception, but also on early return (e.g. `layout_cache`) or `set_name=False` clobbering (that could null out a previous name).

Should be reviewed preferably with white space ignored, since most of the line changes are indentation.

There is a subtle change of `post_process` seeing this cell's name instead of the parent's/None, but I think that's probably desirable?

## Test Plan

Some unit tests. I must admit I asked the clankers to write those tests.


abdelq:
`ty check` failing on CI seems irrelevant to this PR.

sebastian-goeldi (approved):
Thank you @abdelq !

(fixed the ty errors on another branch as it is essentially ty adding new features and not seeing that there is a (crude) model validator doing the right thing)

PR #1041, opened 2026-08-10
Adding max_vertex_count to config

abdelq:
## Summary

This came up again... Last time we added `multi_xy_records`, this time we want `max_vertex_count`...

## Test Plan

N/A


src/kfactory/conf.py
```diff
@@ -265,6 +265,7 @@ class Settings(BaseSettings):
     write_kfactory_settings: bool = True
     """Write kfactory version into the gds/oasis."""
     multi_xy_records: bool = False
+    max_vertex_count: int = 8000
```
sebastian-goeldi, line 268:
```suggestion
    max_vertex_count: int = 4000
```

Strict GDSII standard treats this as a signed integer, so max is 4000 something. So, if we introduce this, we should set it to 4000

sebastian-goeldi (changes requested):

src/kfactory/conf.py
abdelq: I get it, but that would be diverging from KLayout's default https://github.com/KLayout/klayout/blob/31f5ebf340fea0425554fbe618dc750444593854/src/plugins/streamers/gds2/db_plugin/dbGDS2Format.h#L109

sebastian-goeldi: fair enough

sebastian-goeldi (approved):

PR #1045, opened 2026-08-14
Build the types for serialization checks once at import time

abdelq:
## Summary

When using nested types with a lot of elements, profiling showed a lot of time was just spent on the type union per `isinstance` call. It gets costly.

I'd appreciate it if a patch release is made for this one :)


sebastian-goeldi:
hmm, fair. wouldn't it be sufficient to build the union type once outside?

sebastian-goeldi (approved):

PR #1058, opened 2026-08-27
Add per-reference info channel on instances

mpasson:
## Summary

Added an InstanceInfo object to references mimicking the behavior if Info object on cells. This allows a more precise distinction between instances by assigning metadata beyond a specific instance name.

## Test Plan

Usage and GDS roundtrip can be verified with the following script:

```python
"""Cell info vs per-reference (instance) info.

It shows:

1. How to attach ``info`` to a *cell* (shared by every placement).
2. How to attach ``info`` to an *instance* / reference (unique per placement),
   the new per-reference info channel.
3. That both kinds of info survive a GDS write -> read roundtrip, and that
   placing the same cached cell twice keeps a single cell in the layout while
   each instance keeps its own info.
"""

from __future__ import annotations

import tempfile
from pathlib import Path

import kfactory as kf


def build_layout() -> kf.KCLayout:
    """Create a layout with one reusable cell placed twice, each with own info."""
    kcl = kf.KCLayout("INFO_EXAMPLE")
    layer = kcl.layer(1, 0)

    # --- a single, reusable "device" cell -----------------------------------
    # Cell-level info is shared by *every* placement of this cell.
    device = kcl.kcell("mzi_device")
    device.shapes(layer).insert(kf.kdb.DBox(0, 0, 20, 10))
    device.info["component"] = "mzi"
    device.info["length_um"] = 20.0

    # --- a top cell that places the same device twice -----------------------
    top = kcl.kcell("top")

    ref_a = top << device
    ref_b = top << device
    ref_b.dmovey(30)  # move the second placement out of the way

    # Per-reference info: the *same* cached cell, but each placement is
    # measured differently. This is exactly like doing cell.info[...] = ...,
    # only it lives on the instance instead of the cell.
    ref_a.info["measure"] = "spectrum"
    ref_a.info["input_port"] = "o1"
    ref_a.info.wavelength_nm = 1550  # attribute-style assignment also persists

    ref_b.info["measure"] = "power"
    ref_b.info["input_port"] = "o2"

    return kcl


def show(kcl: kf.KCLayout, banner: str) -> None:
    """Pretty-print the cell info and every instance's info under ``top``."""
    print(f"\n{'=' * 70}\n{banner}\n{'=' * 70}")

    top = kcl["top"]
    device = kcl["mzi_device"]

    print(f"cells in layout : {[c.name for c in kcl.each_cell()]}")
    print(f"instances in top: {len(top.insts)} (same cached cell reused)")
    print(f"\ncell 'mzi_device'.info = {dict(device.info)}")

    for inst in top.insts:
        # inst.cell_name is the shared cell; inst.info is unique per placement
        print(f"  instance -> cell={inst.cell_name!r}  info={dict(inst.info)}")


def main() -> None:
    kcl = build_layout()
    show(kcl, "BEFORE WRITE (in memory)")

    with tempfile.TemporaryDirectory() as tmp:
        gds_path = Path(tmp) / "info_example.gds"
        kcl["top"].write(gds_path)
        print(f"\nwrote GDS -> {gds_path}")

        # Read the GDS back into a completely fresh layout.
        kcl_read = kf.KCLayout("INFO_EXAMPLE_READ")
        kcl_read.read(gds_path)
        show(kcl_read, "AFTER GDS ROUNDTRIP (read back from disk)")


if __name__ == "__main__":
    main()
```

### DIFF
--- a/docs/source/concepts/instances.py
+++ b/docs/source/concepts/instances.py
@@ -224,6 +224,46 @@ b30b.flatten()  # merges geometry into parent to close sub-nm gaps
 angled
 
 # %% [markdown]
+# ## Per-instance info
+#
+# A cell's `info` is shared by *every* placement of that cell. When you need
+# metadata that differs **per placement** — e.g. which measurement each copy of a
+# reusable device is used for — attach it to the instance instead via `inst.info`.
+#
+# `inst.info` mirrors `cell.info`: attribute- or item-style access, and the same
+# metadata types (Python scalars/containers and `kdb` shapes). It is stored on the
+# instance's GDS user properties, so different placements of the *same cached cell*
+# carry different info without breaking cell caching, and it survives a GDS/OASIS
+# roundtrip.
+
+# %%
+device = kf.KCell(name="reusable_device")
+device.shapes(kf.kcl.find_layer(L.WG)).insert(kf.kdb.DBox(0, 0, 20, 10))
+device.info["component"] = "mzi"  # shared by every placement
+
+circuit = kf.KCell(name="per_instance_info")
+ref_a = circuit << device
+ref_b = circuit << device
+ref_b.dmovey(30)
+
+# per-placement metadata: same cached cell, different info on each instance
+ref_a.info["measure"] = "spectrum"
+ref_a.info.wavelength_nm = 1550  # attribute-style assignment also persists
+ref_b.info["measure"] = "power"
+
+# a single cell definition is reused, yet each instance keeps its own info
+assert ref_a.cell_index == ref_b.cell_index
+assert dict(ref_a.info) == {"measure": "spectrum", "wavelength_nm": 1550}
+assert dict(ref_b.info) == {"measure": "power"}
+print("cell.info :", dict(device.info))
+print("ref_a.info:", dict(ref_a.info))
+print("ref_b.info:", dict(ref_b.info))
+
+# %% [markdown]
+# > **Note:** `inst.info` lives on the instance's properties, so — like the
+# > instance name — it is discarded when the instance is flattened into its parent.
+
+# %% [markdown]
 # ## Summary
 #
 # | Task | API |
@@ -235,6 +275,7 @@ angled
 # | Regular array | `parent.create_inst(child, na=N, nb=M, a=vec_a, b=vec_b)` |
 # | Access array port | `arr_inst["port_name", col, row]` |
 # | Expose child ports | `parent.add_ports(inst.ports, prefix="…")` |
+# | Per-placement metadata | `inst.info["key"] = value` |
 # | Flatten into parent | `inst.flatten()` |
 
 # %% [markdown]
--- a/src/kfactory/conf.py
+++ b/src/kfactory/conf.py
@@ -64,6 +64,8 @@ class PROPID(IntEnum):
     """Instance name."""
     PURPOSE = 1
     """Instance purpose (e.g. 'routing')."""
+    INFO = 2
+    """Per-instance info dict, stored as a single JSON blob (see ``Instance.info``)."""
 
 
 class PORTDIRECTION(IntEnum):
--- a/src/kfactory/instance.py
+++ b/src/kfactory/instance.py
@@ -25,6 +25,7 @@ from .exceptions import (
 )
 from .geometry import DBUGeometricObject, GeometricObject, UMGeometricObject
 from .port import DPort, Port, ProtoPort
+from .serialization import deserialize_setting, serialize_setting
 from .settings import Info
 
 if TYPE_CHECKING:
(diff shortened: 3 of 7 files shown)


sebastian-goeldi:
```python
ref_b.info["abc"] = kf.kdb.Box(500)
```

makes this fail ;)

sebastian-goeldi:
also, I am kind of against this. The reason is that property fields are limited in size.

e.g. this will make it fail `ref_b.info["really_long_thing"] = "a" * 100000` as the property field size will overflow (maximum afaik is 65KB record length). So, the right way to validate would be to on setting the info stuff to validate that the info can actually be written, which is expensive.


KLayout solves this by smartly splitting long records etc, but this is a bit more complicated and I'd say out of scope for kfactory

mpasson:
Thanks for the feedback, I was a little sloppy on testing :-). 

I am starting to realize this is less straightforward than I though. It can be worked around using string serialization already present, it will not be as native as the cell info.

Regarding validation, in info we already have type checks coercion, so would adding a size check be that expansive on typical objects (strings or short lists)?

Regarding the feature presence, I feel something like this could be useful in photonic design workflow (it certainly be in mine). Right now shared metadata assumes that every placement of each cell is identical in object and purpose, which may not be the case. Doubling a cell just to have different metadata creates more problems that it solves.

I would be open to any alternative suggestion for reference metadata.

sebastian-goeldi:
I think maybe the better approach would actually be to store it in the parent cell and just make sure the instance is named when doing so (otherwise the info changes with position of the instance). I can whip that PR up and the infrastructure would then already be there.

Roughly what I have in mind — `instance.info` becomes a resolved lookup into `parent_cell.instance_info[instance.name]`, with `instance.is_named()` as the gate:

```mermaid
erDiagram
    ProtoTInstance {
        bool is_named "prop_NAME is not None"
        str name "prop_NAME, else derived from trans"
        ProtoTKCell parent_cell FK "-> TKCell"
        Info info "resolved, raises if not is_named()"
    }

    kdb_Instance {
        int prop_NAME "PROPID.NAME = 0, nullable"
        ICplxTrans trans "moving this changes the derived name"
    }

    TKCell {
        int cell_index PK "parent cell, the owner of the rows"
        dict_str_Info instance_info "NEW: instance name -> Info"
    }

    InstanceInfoRow {
        int cell_index PK "FK -> TKCell (parent_cell)"
        str instance_name PK "NOT NULL, requires is_named()"
        Info info "per-placement payload"
    }

    ProtoTInstance ||--|| kdb_Instance    : "wraps _instance"
    ProtoTInstance }o--|| TKCell          : "parent_cell"
    TKCell         ||--o{ InstanceInfoRow : "instance_info[name]"
    ProtoTInstance |o--o| InstanceInfoRow : "info: only when is_named()"
```

The access path, with the naming requirement enforced:

```mermaid
flowchart TD
    A["inst.info"] --> B{"inst.is_named()"}
    B -->|"False (prop_NAME is None)"| E["raise UnnamedInstanceError"]
    B -->|"True"| C["key = inst.name == PROPID.NAME"]
    C --> D["inst.parent_cell._base"]
    D --> F["instance_info: dict[str, Info]"]
    C --> F
    F -->|"hit"| G["Info"]
    F -->|"miss"| H["insert empty Info"]
    H --> G
```

Note the `|o--o|` on the last relationship: an unnamed instance has no row at all, which is the whole point of the `is_named()` gate — the derived `cell_name_x_y[_A..][_M]` fallback in `ProtoTInstance.name` is never usable as a key, since it silently changes the moment you move or rotate the instance.

Two things that fall out of this and need handling in the PR:

- Instance names aren't currently enforced unique per parent cell — `ProtoTInstances._get_inst` just does `next(filter(...))`, so first match wins. For `(cell_index, instance_name)` to actually be a PK, the `name` setter has to reject duplicates within the parent cell.
- Renaming an instance would orphan its row, so `name.setter` needs to migrate the `instance_info` entry along with `PROPID.NAME`.

mpasson:
It fixes the GDS saving problem by only writing Info objects, but the all name book-keeping seems like a workaround for a blocked path.

Wouldn't be possible to keep live instance -> instance_info record in the parent cell and then generate the writable instance_info at write time with whatever name the instance has then? No string book-keeping involved.

In the meantime, i moved forward with my old PR. Unified json serialization for Info and InstanceInfo. GDS export for InstanceInfo uses the json serialization. Info keeps its old GDS path.

Property field size is checked at write time and logged as an error if too big.

Looking forward to your opinion.
[info_roundtrip_example.py](https://github.com/user-attachments/files/31513713/info_roundtrip_example.py)

sebastian-goeldi:
> Wouldn't be possible to keep live instance -> instance_info

You can't really do that easily. The `Instance` objects of kfactory only exist during access time, as they are a wrapper around `kdb.Instance`. You can't really keep references to `kdb.Instance` statically, as it can be reattached to other cells etc. So any book keeping would need to keep this in mind. If the cell is locked, we could do something like we are doing in some cases, i.e. on lock write a static lookup. This should speed up most of the retrievals (but make read etc longer). Also if you delete Instances, the reference can still exist but gets weird errors on access (like `Cell object has been destroyed already` or similar)

So, the short answer is no.

sebastian-goeldi:
I implemented the parent owned infos in https://github.com/gdsfactory/kfactory/pull/1069 . The changes and serialization were trivial to implement.

